PR #5, opened 2019-03-05
Add support for reading partial healsparse maps

erykoff:


healsparse/healSparseMap.py
```diff
@@ -47,12 +47,83 @@ def read(cls, filename, nsideCoverage=None, pixels=None):
...
+                    # This is a table extension
+                    raise RuntimeError("Table extension not yet supported.")
+                    sparseMapSize = hdu.get_nrows()
+
+                nCovPix = sparseMapSize / nFinePerCov - 1
+                covPix, = np.where((sparseMapSize - 2**bitShift) !=
+                                   (np.arange(hp.nside2npix(nsideCoverage), dtype=np.int64) * 2**bitShift) + covIndexMap)
+
+                # Find which pixels are in the coverage map
+                sub = np.clip(np.searchsorted(covPix, pixels), 0, covPix.size - 1)
```
fjaviersanchez, line 99: This is very clever! I don't know if you could just have used `ok  = np.in1d(covPix, pixels)` instead and `sub = covPix[ok]` or is this faster?

fjaviersanchez (approved):
This looks good to me! I have a suggestion for a change but please, feel free to dismiss it if you think it is confusing/wrong. The tests run on my laptop and Travis so this is good to go as soon as you want to merge. Thank you!

healsparse/healSparseMap.py
erykoff: Quite frankly, I didn't know about `np.in1d`.  (There are so many different ways to do things in numpy...)  And in spite of the call sequence being definitely cleaner, it's way slower than the `searchsorted` method.  Huh.

PR #10, opened 2019-03-07
Added downgrade functionality

fjaviersanchez:
I added the downgrade functionality but I made a lot of design choices (e.g, I subbed the `hp.UNSEEN` values by `np.nan` to then calculate the mean). I also decided that this modifies the original `sparseMap`, `nsideSparse` and `bitShift` attributes. Please, let me know if you agree with these choices or you would prefer to expose the method and put a `sparseMap` as input and return a `sparseMap` rather than modifying the original.

### DIFF
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -358,7 +358,7 @@ class HealSparseMap(object):
 
         pass
  
-    def degrade(self, nside_out, reduction='mean', primary=None):
+    def degrade(self, nside_out, reduction='mean'):
         """
         Reduce the resolution, i.e., increase the pixel size
         of a given sparse map
@@ -367,7 +367,6 @@ class HealSparseMap(object):
         ----
         nside_out: `int`, output Nside resolution parameter.
         reduction: `str`, reduction method (mean, median, std, max, min).
-        primary: `str`, in the case of using `np.recarray`, name of the field to use.
         """ 
         
         if self._nsideSparse < nside_out:
@@ -430,5 +429,5 @@ class HealSparseMap(object):
         newIndexMap[self.coverageMask] = np.arange(1, npop_pix + 1) * nFinePerCov
         newIndexMap[:] -= np.arange(hp.nside2npix(self._nsideCoverage), dtype=np.int64) * nFinePerCov
         return HealSparseMap(covIndexMap=newIndexMap, sparseMap=newsparseMap, nsideCoverage=self._nsideCoverage,
-                   nsideSparse=nside_out, primary=primary) 
+                   nsideSparse=nside_out, primary=self._primary) 
 
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -1,5 +1,4 @@
 from __future__ import division, absolute_import, print_function
-
 import numpy as np
 import healpy as hp
 import fitsio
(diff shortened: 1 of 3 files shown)


healsparse/healSparseMap.py
```diff
@@ -262,3 +262,17 @@ def generateHealpixMap(self, nside=None, reduction='mean'):
...
 
+    def degrade(self, nside_out, reduction='mean'):
+        """
+        Reduce the resolution, i.e., increase the pixel size
+        of a given sparse map
+        """
+        try:
+           assert(self._nsideSparse > nside_out)
+        except ValueError as e:
+           print('nside_out should be smaller than nside for the sparseMap')
```
erykoff: This should actually raise an exception.  Try:
```
if self._nsideSparse > nside_out:
    raise ValueError('nside_out should be smaller than nside for the sparseMap')
```

healsparse/healSparseMap.py
```diff
@@ -262,3 +262,17 @@ def generateHealpixMap(self, nside=None, reduction='mean'):
...
+        npop_pix = np.count_nonzero(self.coverageMask)
+        self._sparseMap[self._sparseMap==hp.UNSEEN] = np.nan
+        self._sparseMap = np.nanmean(self._sparseMap.reshape((npop_pix+1, (nside_out//self._nsideCoverage)**2, -1)), axis=2).flatten()
```
erykoff: That's very clever, I didn't know about `np.nanmean` and I bet that takes care of things.  Note that this only works with the `reduction='mean'` option, we should either remove that for the time being, or flag that we will add a (e.g.) median in the future (which will be more expensive of course).

erykoff: So this is going to be have to be updated for the recarray support.  I think that's pretty easy.  If it's a recarray (`self._isRecArray`) then you need to (a) set the sparseMap dtype; (b) loop over fields; but (c) the `UNSEEN` check is on `self._sparseMap[self._primary] == hp.UNSEEN`.
Oh, I just realized that this won't work with the "in-place" replacement, but that's only a couple more lines.

erykoff (changes requested):
So this is going to run into problems with the new format I just pushed.  (hopefully the travis checks catch that!)
I am thinking about the degrade-in-place rather than creating a new map.  I think that the best option is to create a new map, actually, and then if somebody wants to replace it they can do `sparseMap = sparseMap.degrade()` or if they want to hold the old one they do `newSparseMap = sparseMap.degrade()`.

fjaviersanchez:
Thanks! Do you suggest then to write the method into a separate file then (`ud_grade.py`?) or should I keep it in `healSparseMap.py`?

erykoff:
I mean it should be in `HealSparseMap`, it can create one of itself.  So you can call `newMap = oldMap.degrade(nside)` like that.

fjaviersanchez:
since the `hp.UNSEEN` are converted to `np.nan` and the `np.nanmean` or `np.nanmedian` of an array filled with `np.nan` are `np.nan` we could just convert back `np.nan` to `hp.UNSEEN`

fjaviersanchez:
@erykoff Sorry for the delay. Here are the requested changes. I had to separate the recarrays from the regular ndarrays since they were giving trouble with the `np.nan` stuff. I generated a specific test for them including lookups. Please, let me know if you'd like me to restructure anything or have any other requests. Also, I can add more documentation throughout the code in this PR or I can open a new PR dedicated to that. Thanks in advance!

healsparse/healSparseMap.py
```diff
@@ -358,4 +357,84 @@ def generateHealpixMap(self, nside=None, reduction='mean'):
...
+                    dtype.append((key,np.float))
+                else:
+                    dtype.append((key,value[0]))
+            dtype = self._sparseMap.dtype
+            # Allocate new map
+            newsparseMap = np.zeros((npop_pix+1)*nFinePerCov, dtype=dtype)
+            for key, value in newsparseMap.dtype.fields.items():  
+                aux = self._sparseMap[key]
+                aux = aux.reshape((npop_pix+1, (nside_out//self._nsideCoverage)**2, -1))
+                wgt = np.ones_like(aux, dtype=np.float)
```
erykoff: I think this should work (and it does in the tests) but I think we can use the same code as below (`nanmean` etc), we just need to make sure that `aux = self._sparseMap[key].astype(np.float)` and then filling with `NaN`s.  Then we can get the median functionality as well.

erykoff (approved):
Sorry for the long delay, I was travelling.  I think this looks good, with the comment below that might let us add the median functionality (unless I'm missing something, which is very possible).

erykoff (approved):

PR #12, opened 2019-03-12
Add support for building maps, including recarray maps.

erykoff:


healsparse/healSparseMap.py
```diff
@@ -153,24 +157,14 @@ def _readHealSparseFile(filename, pixels=None):
...
-                    # And the rest of the rows
-                    sparseMap[nFinePerCov: rows.size + nFinePerCov] = hdu.read_rows(rows)
+                # It is not 100% sure this is the most efficient way to read in using
+                # fitsio, but it does work.
+                sparseMap = np.zeros((sub.size + 1) * nFinePerCov, dtype=fits['SPARSE'][0:1].dtype)
+                # Read in the overflow bin
+                sparseMap[0: nFinePerCov] = hdu[0: nFinePerCov]
+                # And read in the pixels
+                for i, pix in enumerate(covPix[sub]):
+                    sparseMap[(i + 1)*nFinePerCov: (i + 2)*nFinePerCov] = hdu[covIndexMapTemp[pix]: covIndexMapTemp[pix] + nFinePerCov]
```
fjaviersanchez, line 167: We may want to vectorize this later (I think that the `reshape` trick may work here but it is not trivial). For now it will do I think.

fjaviersanchez (approved):

healsparse/healSparseMap.py
erykoff: Part of the problem is the way that `fitsio` works, I don't know if it's any faster to vectorize and list all the rows vs do the reads in contiguous chunks in the for loop.  But it definitely needs investigation!

PR #13, opened 2019-04-01
generateHealpix

fjaviersanchez:
This PR adds functionality to generate healpix maps from `HealsparseMaps`. I also added some convenience functions to lookup the pixel index for a given position and viceversa, as well as the corresponding tests.


healsparse/healSparseMap.py
```diff
@@ -322,6 +322,26 @@ def getValuePixel(self, pixel, nest=True):
 
         return self._sparseMap[_pix + self._covIndexMap[ipnestCov]]
 
+    def HSPixel2RaDec(self, pixel):
+        """
+        Get the position ra, dec (in degrees) of the center of a certain HealSparse pixel
+        according to its covIndexMap
+        """
+
```
erykoff: Is pixel here the internal pixel index rather than a regular healpix pixel number?  Is this something that we need exposed?

healsparse/healSparseMap.py
```diff
@@ -350,33 +370,69 @@ def coverageMask(self):
...
+            if isinstance(self._sparseMap,int):
+                aux_dtype = np.float #Force to float
+            else:
+                aux_dtype = self._sparseMap.dtype
+            aux = self._sparseMap.astype(aux_dtype)
+        # Create empty HEALPix map
+        hp_map = np.zeros(hp.nside2npix(nside), dtype=aux_dtype) + hp.UNSEEN # remember that empty means UNSEEN!
+        # We generate a HealSparseMap with the requested size before converting to healpix
+        aux_spmap = HealSparseMap(covIndexMap=self._covIndexMap, sparseMap=aux, nsideSparse=self._nsideSparse,
+                                  nsideCoverage=self._nsideCoverage)
```
erykoff: I believe this is only necessary if (a) we are going to degrade the resolution or (b) we have a datatype change (from int to float).  Otherwise, I worry that this uses extra memory.

erykoff (changes requested):
Overall looks good, and the tests are good.  I think we can save a bit of memory (even if the code will get a bit more complex), and I'm not sure about the use case for the convenience functions in `HSPixel2RaDec`.  I'm guessing there's some use that I'm missing, but the documentation will need to be expanded.

healsparse/healSparseMap.py
fjaviersanchez: Hmmm actually, I think that there's no use case for this (I thought that it may have been useful at the time of generating the tests for `generateHealpixMap` but it is not the case). I'll remove them, then.

healsparse/healSparseMap.py
fjaviersanchez: That's a fair comment. I was thinking that, in most cases, this would be a "small" overhead but, I can see that at some point, this is not going to be the case anymore so I changed this. Please, let me know if the changes make sense.

erykoff (approved):

PR #15, opened 2019-04-07
Docs

fjaviersanchez:
Still work in progress but we have some docs available from this branch here: https://healsparse.readthedocs.io/en/latest/

There's also a jupyter notebook associated to this PR.


fjaviersanchez:
I think that this PR is ready to go. The documentation is still sparse but there's something to start with.

The rendered version is here: https://healsparse.readthedocs.io/en/docs/

erykoff (approved):
Sorry for the long delay.  This got lost in the shuffle.  This is really great!  I think it's enough to get it out there, and then we can add more use cases to the tutorial notebook as either (a) we add them or (b) people ask about them (since there are definitely more things that can be done with the code already, and more things that we haven't thought of).

fjaviersanchez:
No worries @erykoff! Thanks for taking a look and yes, I think that as we continue developing we'll see more use cases/get feedback from users.

PR #17, opened 2019-04-09
Add uniform random point generation and tests.

erykoff:
Also add the ability to get a mask of valid pixels (used in random point
generation).


fjaviersanchez (approved):
It looks good to me. Please feel free to merge.

PR #22, opened 2019-05-28
Add map composing operations (sum, product, and, or, etc)

erykoff:
This PR adds operation support to HealSparseMap, and fixes a bunch of small bugs with handling of integer fields.  It also adds support for pure integer HealSparseMap with an integer equivalent to the UNSEEN value.  Single maps can be pulled out of recarray maps with simple key access.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -1,7 +1,7 @@
 Quickstart
 ==========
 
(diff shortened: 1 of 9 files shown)


fjaviersanchez (approved):
It looks great! Thanks a lot. I agree that for the union operations this is probably the best way to go since caching a "master map" can become very expensive memory-wise.

fjaviersanchez:
Can I merge this or are you working on this branch any further @erykoff?

PR #24, opened 2019-06-17
Added plotting routines

fjaviersanchez:
I added the method `hsp_view_map` in `visu_func` to visualize `HealSparseMap` objects. I added some (optional) dependencies in order to support multiple projections. 

I'm still missing an update on the tutorial notebook but, please feel free to take a look at the code and make any suggestions/comments/requests.

Thanks!

### DIFF
--- a/healsparse/visu_func.py
+++ b/healsparse/visu_func.py
@@ -5,14 +5,19 @@ from healsparse.healSparseMap import HealSparseMap
 import healpy as hp
 import matplotlib.pyplot as plt
 from matplotlib.collections import PolyCollection
-import cartopy.crs
 from copy import copy
-from cartopy.mpl.gridliner import LATITUDE_FORMATTER, LONGITUDE_FORMATTER
+# Importing cartopy
+try:
+    import cartopy.crs
+    from cartopy.mpl.gridliner import LATITUDE_FORMATTER, LONGITUDE_FORMATTER
+except ImportError:
+    print('Cartopy is not available, cannot use cartopy projections')i
+    have_cartopy = False
+# Importing shapely
 try:
     import shapely.geometry as sgeom
 except ImportError:
     print('Shapely not found, tick-labels on non-cylindrical projections are not available.')
-
 
 def get_projection(projection_name, central_lon=0):
     """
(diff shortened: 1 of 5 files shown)


fjaviersanchez:
Actually `cartopy` is not optional with the current implementation but I can try to make it optional if necessary.

fjaviersanchez:
I made it optional so the Travis-CI tests would be skipped. I'm planning to add another environment where the tests are performed.

esheldon:
I get this error:

```
File "/home/esheldon/miniconda3/envs/default/lib/python3.7/site-packages/healsparse-0.0.7-py3.7.egg/healsparse/visu_func.py", line 169, in hsp_view_map

AttributeError: 'AxesSubplot' object has no attribute 'set_global'
```

esheldon:
that is because cartopy is not instlalled, so it looks like you probably need to raise an exception if cartopy isn't available, at least with the current code

healsparse/visu_func.py
```diff
@@ -11,7 +11,7 @@
     import cartopy.crs
     from cartopy.mpl.gridliner import LATITUDE_FORMATTER, LONGITUDE_FORMATTER
```
esheldon, line 12: you need to set `have_cartopy = True`

healsparse/visu_func.py
```diff
@@ -0,0 +1,218 @@
...
+    if show_coverage:
+        # Note: We can generalize this to other projections if needed but, for now, we'll stick to mollview
+        plt.axes(ax_cov)
+        hp.mollview(HealSparseMap.coverageMask, title='Coverage Mask', hold=True) #coverageMask is a healpix map
+        
+    if title is not None:
+        fig.suptitle(title, fontsize=20, ha='center')
+    if savename is not None:
+        fig.savefig(savename) 
+    fig.show()
```
esheldon, line 228: need to make `show()` optional, so it won't crash when there is no display

esheldon: Also, I've got cartopy/pyproj installed but nothing shows on the screen, and nothing gets saved to a file

esheldon (changes requested):

esheldon:
Does it make sense to have an option to make the code zoom in on the area that is covered by the map?  For example, I'm implementing small geometric primitives and I'd like to visualize them, but this will I think show the entire sky and a  small circle won't be visible.

erykoff:
I haven't had a chance to look at the code yet (sorry), but @esheldon makes a very good point: the most useful part of `healsparse` is having high resolution over small areas.

fjaviersanchez:
Thanks for the suggestion! The code, as it is right now, can ingest a range of coordinates (`extent`) to show. If there's no extent passed, it will show the full sky. I can change the default behavior to show only the part of the map that is actually covered. Would you say that I should do that change?

esheldon:
`extent` would be enough in some cases, but not for complex maps.  In many cases we won't know what the extent is.

fjaviersanchez:
This is going to go outside of this package.

PR #25, opened 2019-06-18
pixel arrays and masks from geometrical objects

esheldon:
This is the start of that, with a Circle implemented

not really tested yet


esheldon:
TODO:
- add code for poylygon
- add code to combine these geometric primitives into a larger map

esheldon:
test-circle
test-polygon

esheldon:
Time for some bikeshedding:  I dislike the `or_geom` name, we need something better.  `or_primitives`?

esheldon:
test-circles

esheldon:
@fjaviersanchez I tried the visualization stuff with the fixes from @erykoff but it is taking minutes to view the 20 circles in the `test_circles` code

esheldon:
also something actually went wrong with the plot it made
test-vis

esheldon:
mixed geom example
test-mix

esheldon:
need to move tests under the testing directory, although most of these are really just smoke tests at this point.

Need to add a test of the bit values coming out from these complex examples

Need to add a speed test

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
+import numpy as np
+import healpy as hp
+from .healSparseMap import HealSparseMap
+
+
+def make_circles(*, ra, dec, radius, value):
+    """
+    make multiple Circles
```
erykoff, line 8: I think this should have additional doc string to give a particular use case as an example.

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+    """
+    Realize geometry objects in a map
+
+    Parameters
+    ----------
+    geom: geometric primitive or list thereof
+        List of Geom objects, e.g. Circle, Polygon
+    smap: HealSparseMaps
+        The map in which to realize the objects
+    """
```
erykoff, line 82: `type` is missing from the `Parameters`

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+        ----------
+        nside: int
+            Nside for the pixels
+        """
+        raise NotImplementedError('implment get_pixels')
+
+    def get_map(self, *, nside, dtype):
+        """
+        get a healsparse map corresponding to this geometric primitive
+
```
erykoff, line 172: Is this really used in practice, or just in testing?  Because I think that `realize_geom()` is the main way to make a map, and when I was playing around with this I was confused about which I should use.

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+            sentinel=0,
+        )
+        pixels = self.get_pixels(nside=nside)
+        values = self.get_values(size=pixels.size, dtype=dtype)
+        smap.updateValues(pixels, values)
+
+        return smap
+
+    def get_values(self, *, size, dtype):
+        """
```
erykoff, line 198: I guess this is an internal convenience function, should it be `_get_values()`?

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+
+    @property
+    def value(self):
+        """
+        get the value to be used for all pixels in the map
+        """
+        return self._value
+
+    def get_pixels(self, *, nside):
+        """
```
erykoff, line 159: Is this a public function or should it be `_get_pixels()`?

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+        values = np.zeros(size, dtype=dtype)
+        values[:] = self._value
+
+        return values
+
+
+class Circle(GeomBase):
+    def __init__(self, *, ra, dec, radius, value):
+        """
```
erykoff, line 220: Some docstring text here, like "Represent a circle"

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+        return hp.query_polygon(
+            nside,
+            self._vertices,
+            nest=True,
+            inclusive=False,
+        )
+
+
+def test_circle(show=False):
+    ra, dec = 200.0, 0.0
```
erykoff, line 348: Should this test code be here?

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+            show_coverage=False,
+            extent=extent,
+        )
+        """
+        return plt
+    else:
+        return None
+
+def _make_circles(rng, ncircle, ra_range=None, dec_range=None):
+    if ra_range is None:
```
erykoff, line 386: This is also test code, so should be cleaned up / moved / renamed.

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+        ra=ra,
+        dec=dec,
+        radius=radius,
+        value=values,
+    )
+
+    return circles, ra_range, dec_range
+
+def test_circles(show=False, show_mat=False):
+    """
```
erykoff, line 410: Is this in tests now?

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+            smap,
+            savename='/tmp/test.png',
+            show_coverage=False,
+            extent=extent,
+        )
+
+    return plt
+
+def test_box(show=False):
+    ra = [200.0, 200.2, 200.2, 200.0]
```
erykoff, line 481: Is this in tests now?

healsparse/geom.py
```diff
@@ -0,0 +1,668 @@
...
+    dec = [0.0,     0.1,   0.2,   0.25, 0.13]
+    poly = Polygon(
+        ra=ra,
+        dec=dec,
+        value=64,
+    )
+    return poly, ra, dec
+
+def test_polygon(show=False):
+
```
erykoff, line 525: Also, is this in tests now?

erykoff (commented):
Some documentation comments, the tests if they haven't been moved should be moved (and deleted).

healsparse/geom.py
esheldon: yes, but not the visuallization part

maybe leave for now since these are the only examples of fast lightweight visualization

healsparse/geom.py
esheldon: public

healsparse/geom.py
esheldon: public

healsparse/geom.py
esheldon: yes, it is useful because it is so easy to use compared to constructing a map and adding to it

especially nice at the command line

healsparse/geom.py
esheldon: fixed

healsparse/geom.py
esheldon: done

esheldon:
ready to merge?

erykoff:
Looks good.  Thanks!

PR #27, opened 2019-06-19
Add fast random point generator

erykoff:
Add fast (but slightly approximate) random point generator


esheldon (approved):
this looks good to me and it works in my tests

PR #38, opened 2019-09-27
Fix valid pixel computation.

erykoff:


fjaviersanchez (approved):
Thanks! Looks good to me!

PR #46, opened 2020-03-25
Add floating point value support to geometric primitives

erykoff:


fjaviersanchez (approved):
Tests pass and the changes look good to me. Thanks!

PR #49, opened 2020-04-01
Added max/min operations and general ufunc

fjaviersanchez:
This PR adds max/min operations and exposes `_apply_operation` through `ufunc_union` and `ufunc_intersection`.


erykoff (approved):
It looks good.  My only comment is if we want the default to be `np.minimum` and `np.maximum` which propagate NaNs or `np.fmin` and `np.fmax` which ignore NaNs.  Though I don't think we're supposed to have NaNs in the maps so it probably doesn't matter.

fjaviersanchez:
I changed it to `fmin/fmax` just in case (sometimes it's more useful to have `np.nan` than `hp.UNSEEN`)

PR #50, opened 2020-04-01
Add new "wide mask" map type to support > 64 bit bitmasks

erykoff:


fjaviersanchez (approved):
The changes look good to me and I manage to run the tests without any problem. Thanks

PR #51, opened 2020-04-09
Add metadata object to maps to store header information

erykoff:


healsparse/fits_shim.py
```diff
@@ -182,23 +182,31 @@ def _write_filename(filename, c_hdr, s_hdr, cov_index_map, sparse_map):
...
-        return fitsio.FITSHDR(input_header)
+        hdr = fitsio.FITSHDR(metadata)
     else:
-        if input_header is None:
-            return fits.Header()
+        if metadata is None:
+            hdr = fits.Header()
         else:
-            return fits.Header(input_header)
+            hdr = fits.Header(metadata)
```
fjaviersanchez, line 204: Maybe it's worth to invert the order, i.e., get `input_header` first if any, and then allow the option to either overwrite or not if `metadata` shares keys with `input_header`

fjaviersanchez (approved):
Changes look good to me. I'd suggest to allow the user to decide whether they want to overwrite part of the previous header by switching the order in which `metadata` and `input_header` are read and add an overwrite option. I think this is not critical since you could potentially say that `metadata = input_header.toDict()` (or something similar) and then update whichever values you want.

PR #53, opened 2020-04-16
Add install_requires for proper pip installation

erykoff:
Looks like 1.0.1 will be the first version on PyPI...


fjaviersanchez (commented):
Don't we also need `astropy.io.fits`?

erykoff:
So `astropy` is implicit with `healpy`, but I could make it explicit if you like.  (And add it to the requirements as well).  What's your vote?

fjaviersanchez:
Yes, I'd say let's make it explicit just in case `healpy` changes

fjaviersanchez (approved):

PR #54, opened 2020-05-05
Add __getitem__/__setitem__ support to treat sparse maps like full healpix maps

erykoff:


healsparse/healSparseMap.py
```diff
@@ -1202,16 +1204,65 @@ def apply_mask(self, mask_map, mask_bits=None, mask_bit_arr=None, in_place=True)
...
-        values.
-        """
-
-        if not self._is_rec_array:
-            raise TypeError("HealSparseMap is not a recarray map")
-
-        return self.get_single(key, sentinel=None)
+        Get part of a healpix map.
+        """
+        if isinstance(key, str):
```
fjaviersanchez, line 1209: This is very smart!

healsparse/healSparseMap.py
```diff
@@ -1202,16 +1204,65 @@ def apply_mask(self, mask_map, mask_bits=None, mask_bit_arr=None, in_place=True)
...
+            step = key.step if key.step is not None else 1
+            return self.update_values_pix(np.arange(start, stop, step),
+                                          value)
+        elif isinstance(key, np.ndarray):
+            if not is_integer_value(key[0]):
+                raise IndexError("Numpy array indices must be integers for __setitem__")
+            return self.update_values_pix(key, value)
+        elif isinstance(key, list):
+            arr = np.array(key)
+            if not is_integer_value(arr[0]):
```
fjaviersanchez, line 1258: I think any `ndarray` can be also sliced with a list (or `ndarray`) of bools of the same length as the original `ndarray` so that'd be a good addition.

fjaviersanchez (changes requested):
Changes look good and all tests pass in my computer. I'd suggest to have the ability to slice using lists/ndarrays of bools (this is something that at least I use pretty frequently).

fjaviersanchez (approved):
After discussing with @erykoff we decided to just go without the bool slicing. Whoever wants to use it can pass `np.where(bool_list)` to the slicing routines.

PR #56, opened 2020-06-01
Optimize numpy calls in update_values_pix

erykoff:
These small optimizations lead to significantly faster (x2) performance when setting large numbers of pixels (where large is in the millions).


fjaviersanchez (approved):
Everything looks good to me and tests pass in my computer (and it seems to speed tests up quite a bit). Thanks!

PR #57, opened 2020-06-03
Create python-package.yml

erykoff:
Add github actions workflow


fjaviersanchez (approved):
Thanks! It looks good to me. I didn't know about this but it's pretty cool.

PR #63, opened 2020-06-26
Improve performance of valid_pixels retrieval by an order of magnitude

erykoff:
For high-res maps, this improves the time to compute `valid_pixels` by 10-50x, and also greatly decreases the memory footprint.


fjaviersanchez (approved):
Everything looks good to me! Thanks!

PR #64, opened 2020-06-30
Fix cat_healsparse_files for large number of files, and optimize

erykoff:
This branch started as a fix for a bug that triggered when there were a large number of input files.  And then some profiling showed that improving the fits reading would help, but unfortunately it's not much of a difference.


fjaviersanchez (approved):
Looks good to me. Thanks!

Issue #36, opened 2019-07-17
intOnly healspare operations do not check _sentinel dtype (only _sparseMap) 


jacobic:
Hi guys,

Some quick feedback on this great new functionality. I have some healpsparse maps which contain floating values, 1.0, 0.0 and hp.UNSEEN. I want to use the `xorUnion` healsparse operation

https://github.com/LSSTDESC/healsparse/blob/281bc50caa16167de3a7107309ba2a77635a78bd/healsparse/operations.py#L170

so in order to to that I cast my `_sparseMap` to dtype int. However, even with int type maps it is still possible for the numpy ufunc to fail as the sentinel type is not checked even though `intOnly == True`. 

https://github.com/LSSTDESC/healsparse/blob/281bc50caa16167de3a7107309ba2a77635a78bd/healsparse/operations.py#L244

This can be because an int map with a float sentinal passes the `intOnly` requirements of the `_sparseMap` being dtype int but will later fail because the following line results in a dtype float array.

https://github.com/LSSTDESC/healsparse/blob/281bc50caa16167de3a7107309ba2a77635a78bd/healsparse/operations.py#L281  

To summarise I change the following in my maps to ensure the operation works as expected but think some of these check could be done within the `intOnly` check as at the moment it only checks the `_sparseMap`.

```
 m._sparseMap[np.where(m._sparseMap == hp.UNSEEN)[0]] = 0
 m._sparseMap = m._sparseMap.astype(int)
 m._sentinel = 0
```

Keep up the good work! :)

Cheers,
Jacob

erykoff:
Closed with the new `astype` method in #69 .

PR #69, opened 2020-07-06
Add a bunch of small features that have been requested.

erykoff:
- Change default from `lonlat=False` to `lonlat=True`.  This could potentially be a problem for downstream code, but I'm not sure who uses `lonlat=False` (which is I think is @kadrlica 's point).
- Add check for sentinel type when constructing from `healpix_map`.  Note that the sentinel can only be inferred for a floating point map because that is the only type that is defined for a true healpix map.
- Add option to `generate_healpix_map` to generate the map in ring format.
- Add `astype()` conversion method to convert between dtypes, properly handing the sentinel values and valid_pixels.

### DIFF
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -1023,7 +1023,7 @@ class HealSparseMap(object):
 
             self._metadata = metadata
 
-    def generate_healpix_map(self, nside=None, reduction='mean', key=None):
+    def generate_healpix_map(self, nside=None, reduction='mean', key=None, nest=True):
         """
         Generate the associated healpix map
 
@@ -1042,6 +1042,8 @@ class HealSparseMap(object):
         key : `str`
             If the parent HealSparseMap contains recarrays, key selects the
             field that will be transformed into a HEALPix map.
+        nest : `bool`, optional
+            Output healpix map should be in nest format?
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


healsparse/healSparseMap.py
```diff
@@ -71,6 +71,11 @@ def __init__(self, cov_map=None, cov_index_map=None, sparse_map=None, nside_spar
             # this is a healpix_map input
             if sentinel is None:
                 sentinel = hp.UNSEEN
+            if is_integer_value(healpix_map[0]) and not is_integer_value(sentinel):
+                raise ValueError("The sentinel must be set to an integer value with an integer healpix_map")
+            elif not is_integer_value(healpix_map[0]) and is_integer_value(sentinel):
```
fjaviersanchez, line 76: do we need the second check? would python cast it to a float?

healsparse/healSparseMap.py
```diff
@@ -1378,6 +1386,38 @@ def get_single(self, key, sentinel=None, copy=False):
         return HealSparseMap(cov_map=self._cov_map, sparse_map=new_sparse_map,
                              nside_sparse=self._nside_sparse, sentinel=_sentinel)
 
+    def astype(self, dtype, sentinel=None):
```
fjaviersanchez, line 1389: I didn't think of this, it's a nice addition!

fjaviersanchez (approved):
Changes look good and are mostly invisible to the user and test pass. Thanks!

healsparse/healSparseMap.py
erykoff: I've been fairly strict with type checking in the whole code, and I'd like to keep it that way.  But while the harm in allowing implicit conversions from integer to float are also not large (as opposed to the other way), there are certainly cases where I've seen confusion because large integers cannot be converted to float and back.

fjaviersanchez: Thanks! It makes sense.

healsparse/healSparseMap.py
erykoff: This is in response to an ancient issue, #36 .  I don't want people mucking with private variables, but it's clear that it is useful to be able to convert types correctly!

PR #73, opened 2020-07-08
Add support for FITS tile compression of integer and float maps

erykoff:
This adds support for compression of integer maps.  Lossless compression of floating point images is not natively supported by the FITS standard, and thus is not possible at this time.

### DIFF
--- a/healsparse/fits_shim.py
+++ b/healsparse/fits_shim.py
@@ -1,14 +1,15 @@
 import numpy as np
+import mmap
+from .utils import is_integer_value
+# We need this for compression before a newer version of fitsio arrives
+import astropy.io.fits as fits
 
 use_fitsio = False
 try:
     import fitsio
     use_fitsio = True
 except ImportError:
-    try:
-        import astropy.io.fits as fits
-    except ImportError:
-        raise ImportError("Must be able to import either fitsio or astropy.io.fits")
+    pass
 
 
 _image_bitpix2npy = {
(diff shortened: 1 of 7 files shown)


fjaviersanchez (approved):
Looks good! Thanks

erykoff:
I think I'm wrong about the ability to do floating point lossless compression.  I'm going to do a bit more digging, so this might be amended.

erykoff:
Upon further review, it seems that `fitsio` does not currently support writing lossless compressed floating point images (this is a minor problem because we could use `astropy` to write these), but more importantly `fitsio` is broken when _reading_ these images.  This is why I was finding inconsistencies.

erykoff:
@fjaviersanchez Can you take a look at the latest version?  It now can do lossless compression of floating point images as well.

healsparse/fits_shim.py
```diff
@@ -204,15 +206,56 @@ def _write_filename(filename, c_hdr, s_hdr, cov_index_map, sparse_map):
...
+        else:
+            hdu = fits.ImageHDU(data=sparse_map, header=fits.Header())
+
+    for n in s_hdr:
+        hdu.header[n] = s_hdr[n]
+    hdu_list.append(hdu)
+
+    hdu_list.writeto(filename, overwrite=True)
+
+    # When writing a gzip unquantized (lossless) floating point image,
```
fjaviersanchez, line 243: This looks good! Thanks.

fjaviersanchez (approved):
Thanks a lot! This is great.

PR #76, opened 2020-07-14
Issues/72

fjaviersanchez:
Add ability to use AND/OR degrade

### DIFF
--- a/tests/test_degrade.py
+++ b/tests/test_degrade.py
@@ -136,6 +136,34 @@ class DegradeMapTestCase(unittest.TestCase):
         # Check the results
         testing.assert_almost_equal(sparse_map_or._sparse_map, sparse_map_test._sparse_map)
 
+        # Repeat for maxbits > 8
+        sparse_map = healsparse.HealSparseMap.make_empty(nside_coverage, nside_map,
+                                                         WIDE_MASK, wide_mask_maxbits=16)
+        sparse_map_or = healsparse.HealSparseMap.make_empty(nside_coverage, nside_map2,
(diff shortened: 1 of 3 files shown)


erykoff:
Quick comment -- there needs to be support/tests for and/or with straight integer maps, even if they aren't wide masks.

tests/test_degrade.py
```diff
@@ -111,60 +112,122 @@ def test_degrade_map_recarray(self):
...
+        Test HealSparse.degrade AND functionality with WIDE_MASK
         """
-        
+
         nside_coverage = 32
         nside_map = 256
         nside_map2 = 64
         sparse_map = healsparse.HealSparseMap.make_empty(nside_coverage, nside_map,
-                                                 WIDE_MASK, wide_mask_maxbits=7)
+                                                         WIDE_MASK, wide_mask_maxbits=7)
```
erykoff, line 176: The and/or wide_mask tests need two separate tests each, unfortunately, one of them with < 8 bits (like this) and another with > 8 bits.  The one with > 8 bits needs to ensure that some pixels have a low bit set, some pixels have a high bit set, and some have both.  (I have learned this the hard way!)

erykoff (changes requested):
See comments about additional wide_mask tests.  Otherwise looks good.

tests/test_degrade.py
fjaviersanchez: Great suggestion, this allowed me to catch a couple of bugs.

erykoff (approved):
Looks good, I think the tests have good coverage!

PR #77, opened 2020-07-22
Fix bug where setting wide_mask would get confused with certain sizes.

erykoff:


fjaviersanchez (approved):
Looks good. Thanks for fixing this so quickly!

PR #78, opened 2020-08-07
Add fix for compressed file writing when mmap is not implemented

erykoff:
Some filesystems (notably on nersc compute nodes) do not support the mmap trick to patch compressed file headers.  This checks for that and adds a workaround.


fjaviersanchez (approved):
The changes look good. I didn't know this was an issue at NERSC. Thanks for the changes and the heads up

PR #84, opened 2020-08-25
Update doctstrings for degrade

fjaviersanchez:
Currently the docs don't show the ability to degrade with sum, product. This PR adds this.

### DIFF
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -1120,8 +1120,9 @@ class HealSparseMap(object):
             sparsemap nside_sparse
         reduction : `str`
             If a change in resolution is requested, this controls the method to
-            reduce the map computing the mean, median, std, max, min, sum or
-            product of the neighboring pixels to compute the "degraded" map.
+            reduce the map computing the "mean", "median", "std", "max", "min", 
+            "sum" or "prod" (product)  of the neighboring pixels to compute the
+            "degraded" map.
         key : `str`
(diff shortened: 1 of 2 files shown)


healsparse/healSparseMap.py
```diff
@@ -1120,8 +1120,8 @@ def generate_healpix_map(self, nside=None, reduction='mean', key=None, nest=True
             sparsemap nside_sparse
         reduction : `str`
             If a change in resolution is requested, this controls the method to
-            reduce the map computing the mean, median, std, max or min of the
-            neighboring pixels to compute the "degraded" map.
+            reduce the map computing the mean, median, std, max, min, sum or
+            product of the neighboring pixels to compute the "degraded" map.
```
erykoff: I think that we should put quotes around the reduction options.  At the same time it is `"prod"` not `"product"`.  Maybe should be `computing the "mean", "median", "std" (standard deviation), "max", "min", "sum", or "prod" (product) of `...

erykoff (changes requested):
The docstring for `generate_healpix_map` should be made more explicit: https://github.com/LSSTDESC/healsparse/blob/74f5e790683cbe9eaf1061ab262f96261c96537f/healsparse/healSparseMap.py#L1121
Also, the RTD docs should be updated: https://github.com/LSSTDESC/healsparse/blob/master/docs/basic_interface.rst

erykoff (approved):
Looks good!

Issue #61, opened 2020-06-18
Add option to "or" overlapping pixels when concatenating integer maps


erykoff:
Currently, when concatenating maps with `cat_healsparse_files` the only option is to overwrite values when two maps have the same pixel defined.  When concatenating integer mask maps, it is often useful to be able to "or" together the overlaps (which allows, for example, efficient parallelization of rendering of star masks).

fjaviersanchez:
Closed by #88

PR #88, opened 2020-11-05
Added or addition to overlapping integer maps

fjaviersanchez:
This PR tries to address issue #61. If the `check_overlap` option is triggered and the maps is integer the overlapping pixels are `or` concatenated.


erykoff (changes requested):
Looks good, except I think we need to have this as an explicit option (unsure if the default should be true or false).  Not every integer map is a bit-mask that is suitable for or-ing; an nexp map for example you wouldn't want this operation, but you still might want to check for overlapping pixels and Raise if there are any.

fjaviersanchez:
Thanks! Sounds good. I was thinking about this but I didn't really have a good option for the option name or how to make clear what it does... Let me please try to think of something :)

erykoff (approved):
Looks good!

PR #89, opened 2020-11-13
Update datatype casting to work on osx-arm64.

erykoff:
The float to uint conversion used to create a test comparison doesn't work the same on arm.  Luckily, this construction is only used in our test code.


fjaviersanchez (approved):
Already updated to Big Sur?! haha, so fast. Thanks!

erykoff:
This isn't for Big Sur OS, this is for the new Apple Silicon!  We want people to be able to say `conda install healsparse` from conda-forge next week, and have a good experience.  And they will!

PR #91, opened 2020-11-19
Fixed bug

fjaviersanchez:
This PR tries to fix issue 90


erykoff:
I think we also want to explicitly add this to the tests.

erykoff (approved):
Looks good!

Issue #92, opened 2020-11-30
Implement weighted mean degradation


fjaviersanchez:
This would be useful to define means for deprojection/survey property maps weighted by the variance in each pixel.

fjaviersanchez:
Addressed by #93

PR #93, opened 2020-12-03
Added weighted degrade

fjaviersanchez:
This PR tries to address #92.


docs/basic_interface.rst
```diff
@@ -65,7 +65,7 @@ A :code:`HealSparseMap` has the concept of "valid pixels", the pixels over which
     >>> [0.00932548 ... 0.81134431]
 
 
-You can convert a :code:`HealSparseMap` to a :code:`healpy` map (:code:`numpy` array) either by using a full slice (:code:`[:]`) or with the :code:`generate_healpix_map()` method.  Do watch out, at high resolution this can blow away your memory!  In these cases, :code:`generate_healpix_map()` can degrade the map before conversion, using a reduction function (over valid pixels) of your choosing, including :code:`mean`, :code:`median`, :code:`std`, :code:`max`, :code:`min`, :code: `and`, :code: `or`, :code: `sum`, and :code: `prod` (product).
+You can convert a :code:`HealSparseMap` to a :code:`healpy` map (:code:`numpy` array) either by using a full slice (:code:`[:]`) or with the :code:`generate_healpix_map()` method.  Do watch out, at high resolution this can blow away your memory!  In these cases, :code:`generate_healpix_map()` can degrade the map before conversion, using a reduction function (over valid pixels) of your choosing, including :code:`mean`, :code:`median`, :code:`std`, :code:`max`, :code:`min`, :code: `and`, :code: `or`, :code: `sum`, :code: `prod` (product), and :code: `wavg` (weighted average).
```
erykoff: I think this should be called `wmean` instead of `wavg` to (a) be consistent with the `mean` terminology, and (b) be consistent with `suprême`.

healsparse/healSparseMap.py
```diff
@@ -1236,7 +1236,10 @@ def degrade(self, nside_out, reduction='mean'):
         nside_out : `int`
            Output Nside resolution parameter.
         reduction : `str`
-           Reduction method (mean, median, std, max, min, and, or, sum, prod).
+           Reduction method (mean, median, std, max, min, and, or, sum, prod, wavg).
```
erykoff: wavg -> wmean.

healsparse/healSparseMap.py
```diff
@@ -1257,6 +1260,8 @@ def degrade(self, nside_out, reduction='mean'):
                 raise NotImplementedError('Cannot degrade a wide_mask map with this \
                 reduction operation, try and/or')
             else:
+                if weights is not None:
+                    raise Warning('Ignoring weights')
```
erykoff: This warning should say why we're ignoring the weights (can't use for wide mask).

healsparse/healSparseMap.py
```diff
@@ -1267,6 +1272,22 @@ def degrade(self, nside_out, reduction='mean'):
             dtype = []
             sentinel_out = hp.UNSEEN
             # We should avoid integers
+            if weights is not None:
+                if reduction == 'wavg':
```
erykoff: wavg -> wmean

healsparse/healSparseMap.py
```diff
@@ -1267,6 +1272,22 @@ def degrade(self, nside_out, reduction='mean'):
...
+                    if bad_map:
+                        raise ValueError('weights dimensions should be the same as this maps')
+                    else:
+                        weights = weights._sparse_map
+                        # Set to zero weight those pixels that are not observed
+                        weights[self._sparse_map[self._primary] == self._sentinel] = 0.
+                        weights = weights.reshape((npop_pix + 1,
+                                                  (nside_out//self.nside_coverage)**2, -1))
+                else:
+                    raise Warning('Only `wavg` uses weights')
```
erykoff: "Not using weights because only wmean reduction uses weights."

healsparse/healSparseMap.py
```diff
@@ -1279,13 +1300,16 @@ def degrade(self, nside_out, reduction='mean'):
...
-                aux = reduce_array(aux, reduction=reduction)
+                aux = reduce_array(aux, reduction=reduction, weights=weights)
                 # Transform back to sentinel value
                 aux[np.isnan(aux)] = sentinel_out
                 new_sparse_map[key] = aux
 
         # Work with int array and ndarray
         elif (issubclass(self._sparse_map.dtype.type, np.integer)) and (reduction in ['and', 'or']):
+            if weights is not None:
+                raise Warning('Ignoring weights')
```
erykoff: "Weights will not be used with (reduction) mode"

healsparse/healSparseMap.py
```diff
@@ -1294,11 +1318,24 @@ def degrade(self, nside_out, reduction='mean'):
                 aux_dtype = np.float64
             else:
                 aux_dtype = self._sparse_map.dtype
+            if weights is not None:
+                if reduction == 'wavg':
```
erykoff: wavg -> wmean

healsparse/healSparseMap.py
```diff
@@ -1294,11 +1318,24 @@ def degrade(self, nside_out, reduction='mean'):
...
+                    bad_map = ((weights.nside_sparse != self.nside_sparse) or
+                               (weights.nside_coverage != self.nside_coverage) or
+                               (not np.allclose(weights.valid_pixels, self.valid_pixels)))
```
erykoff: Why is this guard not the same as above (`weights.valid_pixels == self.valid_pixels`)?

healsparse/utils.py
```diff
@@ -48,6 +48,15 @@ def reduce_array(x, reduction='mean', axis=2):
             ret = np.bitwise_and.reduce(x, axis=axis).ravel()
         elif reduction == 'or':
             ret = np.bitwise_or.reduce(x, axis=axis).ravel()
+        elif reduction == 'wavg':
```
erykoff: wavg -> wmean

tests/test_degrade.py
```diff
@@ -380,6 +380,35 @@ def test_degrade_map_int_sum(self):
...
+    def test_degrade_map_wavg(self):
+        """
+        Test HealSparse.degrade functionality with float quantities
+        """
+        random.seed(12345)
+        nside_coverage = 32
+        nside_map = 1024
+        nside_new = 256
+        full_map = random.random(hp.nside2npix(nside_map))
+        weights = 0.25*np.ones_like(full_map)
```
erykoff: I think we need to test with weights that vary.  If they are all the same, this is completely equivalent to an unweighted mean.  And that's what tested below, with the ud_grade (unweighted).  And I guess this tests that the code path is valid, it doesn't actually test that this is doing the weights correctly (e.g., we could change the wmean to mean and it'll give the same answer and still pass tests.)

tests/test_degrade.py
```diff
@@ -380,6 +380,35 @@ def test_degrade_map_int_sum(self):
...
+        deg_map = hp.ud_grade(full_map, nside_out=nside_new, order_in='NESTED', order_out='NESTED')
+
+        # Degrade sparse map and compare to original
+
+        new_map = sparse_map.degrade(nside_out=nside_new, reduction='wavg', weights=weights)
+
+        # Test the coverage map generation and lookup
+
+        testing.assert_almost_equal(deg_map, new_map.generate_healpix_map())
+
```
erykoff, line 411: I'm wondering if we need another test looking for the warnings?  This is less important.

erykoff (changes requested):
Some suggested updates on terminology/warnings, and I think the tests need to be broadened.  Thanks!

healsparse/healSparseMap.py
fjaviersanchez: Oh, that's a bug! `weights.valid_pixels != self.valid_pixels` doesn't work (it throws a `ValueError` because I'm comparing a bitwise quantity with an ndarray of bools. I just put `weights.valid_pixels.all() != self.valid_pixels.all()`

tests/test_degrade.py
fjaviersanchez: right! This was me being lazy 😛.

healsparse/healSparseMap.py
erykoff: That doesn't make sense ... `valid_pixels` should be an array of integers.  How is it not?

erykoff: Oh, the problem is that you need `np.array_equal(weights.valid_pixels, self.valid_pixels)`.

erykoff: (The `weights.valid_pixels.all() != self.valid_pixels.all()` doesn't work either, by the way.).

erykoff (approved):
Removing that debugging line and it looks great!

fjaviersanchez:
Sorry, I missed it! Now the `print` is out.

PR #94, opened 2020-12-04
Add ability to degrade-on-read, with or without weights.

erykoff:


healsparse/healSparseMap.py
```diff
@@ -320,6 +348,11 @@ def _read_healsparse_file(filename, pixels=None):
            Name of the file to read.
         pixels : `list`, optional
            List of integer pixels from the coverage map
+           degrade_nside : `int`, optional
+           Degrade map to this nside on read.  None means leave as-is.
+           reduction : `str`, optional
+           Reduction method with degrade-on-read.
+           (mean, median, std, max, min, and, or, sum, prod).
```
fjaviersanchez: Is `wmean` missing intentionally?

fjaviersanchez (approved):
Thanks a lot! The changes are great and tests pass for me.

healsparse/healSparseMap.py
erykoff: These docs should not have been there at all, I cleaned that up.  Thanks for noticing!

PR #95, opened 2021-01-05
Fix degrade operation on maps with out-of-order coverage map.

erykoff:


fjaviersanchez (approved):
Good catch! Thank you!

PR #96, opened 2021-02-05
Replace np.bool with np.bool_ and np.float with np.float_

erykoff:
Note that np.bool, np.float are deprecated with numpy 1.20.


fjaviersanchez (changes requested):
Thanks @erykoff. There are still a couple of rogue `np.float` in here: https://github.com/LSSTDESC/healsparse/blob/master/tutorial/quickstart.ipynb

fjaviersanchez (approved):
Awesome! Thanks :)

PR #97, opened 2021-02-24
Fix crash when trying to read coverage pixel larger than coverage.

erykoff:
There was an off-by-one error on a bounds check that was letting out-of-bounds values in.  This fixes that bug and adds an explicit test.


healsparse/healSparseMap.py
```diff
@@ -385,7 +385,7 @@ def _read_healsparse_file(filename, pixels=None):
 
             # Which pixels are in the coverage map?
             cov_pix, = np.where(cov_map.coverage_mask)
-            sub = np.clip(np.searchsorted(cov_pix, _pixels), 0, cov_pix.size)
+            sub = np.clip(np.searchsorted(cov_pix, _pixels), 0, cov_pix.size - 1)
```
fjaviersanchez, line 388: Nice catch!

fjaviersanchez (approved):
Changes look good to me and tests pass for me. Thank you!

PR #98, opened 2021-03-31
Fix multiple bit reset bugs and add appropriate tests.

erykoff:
The bit clear code was broken when (a) more than one bit was cleared or (b) the wide mask was larger than 8 bits.  This adds the appropriate tests and fixes the bug.


erykoff:
Turns out that this clear bug was hiding another significant bug in the index ordering for degrading wide masks (because the error in the clear had the same effect as the degrade error so our tests weren't catching either error).

fjaviersanchez (approved):
Thank you! Changes look good to me.

Issue #99, opened 2021-05-13
Trying to degrade a map below nside_coverage gives unhelpful error.


erykoff:
Right now if you try to run, e.g., `m.degrade(8)` on a map with `nside_coverage=32`, then you get an unhelpful error `ValueError: cannot reshape array`.  This should either (a) raise with a more helpful error, or (possibly; this might be slow and use too much memory?) (b) remake the map with lower nside_coverage and then degrade.

fjaviersanchez:
Closed by #100

PR #100, opened 2021-05-14
Issues/99

fjaviersanchez:
This PR addresses #99 by generating a new map with lower `nside_coverage` to accomodate the requested `nside_out`. It raises a `ResourceWarning` so the user is aware that they may fill their memory pretty quickly. This also fixes another issue discovered by @erykoff for `__setitem__` using integers.

### DIFF
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -7,6 +7,7 @@ from .healSparseCoverage import HealSparseCoverage
 from .utils import reduce_array, check_sentinel, _get_field_and_bitval, WIDE_NBIT, WIDE_MASK
 from .utils import is_integer_value, _compute_bitshift
 from .fits_shim import HealSparseFits, _make_header, _write_filename
+import warnings
 
 
 class HealSparseMap(object):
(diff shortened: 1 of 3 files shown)


healsparse/healSparseMap.py
```diff
@@ -1467,9 +1467,44 @@ def valid_pixels_pos(self, lonlat=True, return_pixels=False):
...
+        if nside_out < self.nside_coverage:
+            # The way we do the reduction requires nside_out to be >= nside_coverage
+            # we allocate a new map with the required nside_out
+            # CAUTION: This may require a lot of memory!!
+            raise ResourceWarning("`nside_out` < `nside_coverage`. \
+                                  Allocating new map with nside_coverage=nside_out")
+            sparse_map_out = HealSparseMap.make_empty_like(self,
+                                                           nside_coverage=nside_out)
+            sparse_map_out.update_values_pix(self.valid_pixels,
+                                             self.get_values_pix(self.valid_pixels))
```
erykoff: This is a bit prettier with `sparse_map_out[self.valid_pixels] = self[self.valid_pixels]` or even a bit faster to do this to prevent 2 lookups of valid_pixels.
```
valid_pixels = self.valid_pixels
sparse_map_out[valid_pixels] = self[valid_pixels]
```

healsparse/healSparseMap.py
```diff
@@ -1467,9 +1467,44 @@ def valid_pixels_pos(self, lonlat=True, return_pixels=False):
...
+            sparse_map_out = sparse_map_out._degrade(nside_out, reduction=reduction, weights=weights)
```
erykoff: The weight map has to be similarly converted to a different nside_coverage; we require that the weights and the map must match (see code below).

tests/test_degrade.py
```diff
@@ -537,6 +537,96 @@ def test_degrade_widemask_wmean(self):
...
+            if method == 'wmean':
+                wgt = weights
+            else:
+                wgt = None
+            sparse_map = healsparse.HealSparseMap.make_empty(nside_coverage, nside_map, dtype=np.float64)
+            sparse_map2 = healsparse.HealSparseMap.make_empty(nside_out, nside_map, dtype=np.float64)
+            sparse_map.update_values_pix(pxnums, pxvalues)
+            sparse_map2.update_values_pix(pxnums, pxvalues)
+            # Check that a warning is raised
+            # testing.assert_raises(ResourceWarning, sparse_map.degrade,
```
erykoff: Remove the commented code.

tests/test_degrade.py
```diff
@@ -537,6 +537,96 @@ def test_degrade_widemask_wmean(self):
...
+            #                      nside_out, reduction=method, weights=weights)
+            with testing.assert_raises(ResourceWarning):
+                sparse_map = sparse_map.degrade(nside_out, reduction=method, weights=wgt)
```
erykoff, line 568: I'm not sure how this could work ... the weights aren't a healsparse map but https://github.com/LSSTDESC/healsparse/blob/master/healsparse/healSparseMap.py#L1503-L1504

erykoff (commented):
We should also have a test for setting a map with a numpy integer pixel.
I'm also confused about the weights, how those tests are passing.

healsparse/healSparseMap.py
fjaviersanchez: Apparently the `assert_raises` was trumping everything else. The degrade code wasn't even being called.

erykoff (approved):
Looks good! Thanks!

Issue #101, opened 2021-06-04
Warn against poor performance with very large nside_coverage


erykoff:
@rgruendl has pointed out that things did not work well when he (as a novice user) had tried `nside_coverage=1024`.  And in fact, this is not surprising!

In my usage I have not found anything greater than `nside_coverage>128` to be useful.

I will do a couple of benchmark tests to confirm where things really do start struggling, and add a warning if a higher `nside_coverage` is specified to protect against these sorts of mistakes.

fjaviersanchez:
Closed by #103

PR #103, opened 2021-06-13
Added warning when using large nside_coverage

fjaviersanchez:
This PR addresses #101, adding a warning with `nside_coverage >= 128`

### DIFF
--- a/healsparse/healSparseCoverage.py
+++ b/healsparse/healSparseCoverage.py
@@ -77,7 +77,7 @@ class HealSparseCoverage(object):
            HealSparseCoverage from file
         """
         if nside_coverage >= 128:
-            warnings.warn('Using `nside_coverage` >= 128 may allocate large arrays in memory!') 
+            warnings.warn('Using `nside_coverage` >= 128 may allocate large arrays in memory!')
         bit_shift = _compute_bitshift(nside_coverage, nside_sparse)
         nfine_per_cov = 2**bit_shift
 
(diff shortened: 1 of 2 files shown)


healsparse/healSparseCoverage.py
```diff
@@ -75,6 +76,8 @@ def make_empty(cls, nside_coverage, nside_sparse):
         healSparseCoverage : `HealSparseCoverage`
            HealSparseCoverage from file
         """
+        if nside_coverage >= 128:
+            warnings.warn('Using `nside_coverage` >= 128 may allocate large arrays in memory!')
```
erykoff: First of all, I'm not 100% sure if 128 is the right nside to do a warning.  I think we might need to benchmark it.  However, anecdotally I can say that I have found nside=128 useful in some contexts (and not overwhelming), so these warnings should be `> 128` and not `>= 128`.  I also wonder if we should have a flag to suppress the warning if the user knows what they're doing (or they can always wrap in a `warnings` context, which is more flexible).
Also, I think this should be a `ResourceWarning` which is "for warnings related to resource usage."  Which I think this is, but maybe `UserWarning` is appropriate?  This I'm less sure of.

erykoff: The other problem, not just the large arrays, is that the performance is going to suffer if nside coverage gets too large.

erykoff (changes requested):

erykoff (approved):
Looks good.

PR #106, opened 2021-06-17
Improve handling of non-unique pixels in update operations.

erykoff:


healsparse/_version.py
```diff
@@ -1 +1 @@
-__version__ = '1.2.8'
+__version__ = '1.2.9'
```
beckermr: bumping to 1.3.0 since there is a new feature would be sensible, though not required.

healsparse/healSparseMap.py
```diff
@@ -771,19 +771,57 @@ def _reserve_cov_pix(self, new_cov_pix):
...
         self._sparse_map[oldsize:] = self._sparse_map[0]
 
+    def update_values_pos(self, theta_or_ra, phi_or_dec, values,
+                          lonlat=True, operation='replace'):
+        """
+        Update the values in the sparsemap for a list of positions.
+
+        Parameters
+        ----------
+        theta_or_ra : `float`, array-like
```
fjaviersanchez: Maybe write `ra_or_theta` (since the `lonlat` option is by default `True`?

fjaviersanchez (approved):
Thanks @erykoff! The PR looks good to me. I have a minor documentation comment, but it looks good to go.

PR #113, opened 2021-07-28
Change versioning to setuptools_scm_git_archive to sync with tags.

erykoff:
I recently learned from @beckermr for this one simple trick to get the versioning to sync with git tags.  Now versions will be auto-generated with the git ref, and if we push a new tag then it gets incorporated into the package automatically (and a release is made with that tag).


erykoff:
For a bit more info, I was just copying what was done here: https://github.com/beckermr/pizza-cutter and looking at the docs here: https://pypi.org/project/setuptools-scm-git-archive/

beckermr (approved):

healsparse/healSparseMap.py
```diff
@@ -164,7 +164,7 @@ def read(cls, filename, nside_coverage=None, pixels=None, header=False,
                     row = fits.read_ext_data(1, row_range=[0, 1])
                     dtype = row[0][0][0].dtype.type
 
-            healpix_map = hp.read_map(filename, nest=True, verbose=False, dtype=dtype)
+            healpix_map = hp.read_map(filename, nest=True, dtype=dtype)
```
fjaviersanchez, line 167: Was this change intentional?

fjaviersanchez (approved):
Thanks @erykoff! Looks good to me

healsparse/healSparseMap.py
erykoff: Yes, this was a separate commit.  I removed it because the `verbose` keyword has been deprecated and does nothing.

PR #116, opened 2021-10-01
Add parquet file serialization.

erykoff:


docs/filespec.rst
```diff
@@ -103,20 +153,105 @@ If the sparse map is a wide mask, it must contain:
...
+
+Additional metadata from the map is stored as a FITS header string (for compatibility with the FITS serialization) such that:
+* :code:`'healsparse::header'`: :code:`header_string`
+
+Parquet Coverage Map
+--------------------
+
+The coverage map is a Parquet file with the name :code:`_coverage.parquet`, stored in the dataset directory.
+The coverage map has two columns:
+* :code:`cov_pix`: Valid coverage pixeld (:code:`nside = nside_coverage`) for the sparse map.
```
fjaviersanchez: pixeld -> pixels?

docs/filespec.rst
```diff
@@ -103,20 +153,105 @@ If the sparse map is a wide mask, it must contain:
...
+* :code:`row_group`: The row group index within the appropriate i/o pixel file to find the sparse data for the given coverage map.
+
+Parquet Map Files
+-----------------
+
+Each sparse map Parquet file covers one i/o pixel.
+The name of each file is :code:`iopix=###/###.parquet`, where :code:`###` is a zero-padded three digit number for the given i/o pixel.
```
fjaviersanchez, line 217: what happens if you were to use `nside_io >= 16`? I think you would need more than 3 digits?

docs/filespec.rst
```diff
@@ -103,20 +153,105 @@ If the sparse map is a wide mask, it must contain:
...
+* :code:`cov_pix`: :code:`int32`
+* :code:`sparse`: Datatype of the sparse image data.
+
+The :code:`cov_pix` gives the coverage pixel of the data, and is redundant with the data in :code:`_coverage.parquet`.
+It compresses very efficiently, and can be used to reconstruct the :code:`_coverage.parquet` from the data files if necessary.
+
+The :code:`sparse` column has the sparse map data (with sparse map image datatype).
+
+Unlike the FITS serialization, the initial "overflow" coverage pixel is not serialized.
+Instead, on read this is filled in with the :code:`sentinel` value from the Parquet metadata.
```
fjaviersanchez, line 239: This is very cool!

docs/filespec.rst
```diff
@@ -103,20 +153,105 @@ If the sparse map is a wide mask, it must contain:
...
+
+**Sparse Map Wide Mask**
+
+If the sparse map is a wide mask map, the schema is the same as for a regular sparse map image.
+In this case, as with the FITS serialization, the sparse map is stored as a flattened version of the in-memry :code:`wide_mask_width * npix` array.
```
fjaviersanchez: in-memry -> in-memory

healsparse/io_coverage_parquet.py
```diff
@@ -0,0 +1,49 @@
...
+    cov_fname = os.path.join(filepath, '_coverage.parquet')
+    if not os.path.isfile(cov_fname):
+        # Note that this could be reconstructed from the information in the file
+        # inefficiently.  This feature could be added in the future.
+        raise RuntimeError("Filepath %s is missing coverage map %s" % (filepath, cov_fname))
+
+    nside_sparse = int(md['healsparse::nside_sparse'])
+    nside_coverage = int(md['healsparse::nside_coverage'])
+
+    cov_tab = parquet.read_table(cov_fname, use_threads=False, columns=['cov_pix'])
```
fjaviersanchez: I'm thinking that there are situations where you may want to change `use_threads=True`, so maybe it's worth exposing the option (I admit that I'm thinking of a situation where probably you would be fine using the FITS read/write approach).

healsparse/io_map_parquet.py
```diff
@@ -0,0 +1,314 @@
...
+    """
+    if os.path.isfile(filepath) or os.path.isdir(filepath):
+        raise RuntimeError("Filepath %s exists and clobber is not supported." % (filepath))
+
+    # Make the path
+    os.makedirs(filepath)
+
+    # Create the nside_io paths
+    cov_mask = hsp_map.coverage_mask
+
```
fjaviersanchez, line 234: I am not sure if this is going to be an issue, but I think it'd be worth checking thta `nside_io < nside_coverage`

healsparse/io_map_parquet.py
```diff
@@ -0,0 +1,314 @@
...
+    cov_fname = os.path.join(filepath, '_coverage.parquet')
+    if not os.path.isfile(cov_fname):
+        # Note that this could be reconstructed from the information in the file
+        # inefficiently.  This feature could be added in the future.
+        raise RuntimeError("Filepath %s is missing coverage map %s" % (filepath, cov_fname))
+
+    nside_sparse = int(md['healsparse::nside_sparse'])
+    nside_coverage = int(md['healsparse::nside_coverage'])
+    nside_io = int(md['healsparse::nside_io'])
+    bitshift_io = _compute_bitshift(nside_io, nside_coverage)
```
fjaviersanchez, line 78: Not sure if we need to guard against `nside_io < nside_cov` here.

fjaviersanchez (changes requested):
Thanks a lot! This was a lot of work, so thanks a lot! My only request is to add some guard to prevent that `nside_io > nside_cov`, and maybe add a test where it explicitly tries to create a map with this configuration. Otherwise it looks great. Thanks again!

docs/filespec.rst
erykoff: I believe this is inefficient even for the highest resolution maps.  I'll add a check to limit a max nside_io as well.

healsparse/io_map_parquet.py
erykoff: I don't think we can guard against all forms of corrupted files, because such a thing could not have been written in the first place.

fjaviersanchez (approved):
Changes look great. Thanks!

Issue #117, opened 2022-01-13
Create `n_valid` property


erykoff:
@esheldon notes it is often convenient to compute the number of valid pixels in a map without generating the list of valid pixels (faster and less memory).  This can be done in the simple case with `n_valid = np.sum(self._sparse_map != self._sentinel)` with slight modifications for recarrays and wide masks.

The map area coverage should also use this routine.

PR #118, opened 2022-01-13
Add n_valid property

erykoff:
This addresses #117 in a more memory efficient way than rendering valid_pixels and counting them.


fjaviersanchez (approved):
Thank you @erykoff. This is much more efficient and a great addition.

PR #120, opened 2022-01-14
Add ability to upgrade maps as well as degrade them.

erykoff:
This PR also adds the ability to get values out of a map using higher nside pixels (for convenience), as well as the ability to get a copy of the map from a single coverage pixel, and an easy way of iterating over all coverage pixels.  For large high-res maps this functionality will allow downstream code to have less memory overhead.

### DIFF
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -664,7 +664,7 @@ class HealSparseMap(object):
         """
         Get the map value for a set of pixels.
 
-        This routine will optionally convert the nside of the input pixels
+        This routine will optionally convert from a higher resolution nside
         to the nside of the sparse map.
 
         Parameters
@@ -677,6 +677,7 @@ class HealSparseMap(object):
             Return mask of True/False instead of values
         nside : `int`, optional
             nside of pixels, if different from native.
+            Must be greater than the native nside.
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


healsparse/healSparseMap.py
```diff
@@ -1300,18 +1313,20 @@ def _degrade(self, nside_out, reduction='mean', weights=None):
 
     def degrade(self, nside_out, reduction='mean', weights=None):
```
fjaviersanchez, line 1314: Do we want to change the name of the function to `ud_grade`?

fjaviersanchez: I guess we want to keep `degrade` for legacy purposes, but maybe also create a `ud_grade` to follow healpy's convention?

fjaviersanchez (approved):
Thanks! The changes look good to me. I have a couple of short questions that can be addressed here, and a longer question that I'll ask on Slack :)

healsparse/healSparseMap.py
erykoff: At first I had a separate `upgrade` method, and I could easily be persuaded to use that directly.  I really dislike the `ud_grade` name in healpy, so I personally would vote against that.

erykoff: upgrade and degrade are now separated.

fjaviersanchez (approved):
Looks good! Thanks

Issue #122, opened 2022-01-28
Boolean maps can't be filled with booleans


kadrlica:
Here's a fun one...
```
>>> map_bool = hsp.HealSparseMap.make_empty(32, 16384, np.bool)
>>> map_bool
HealSparseMap: nside_coverage = 32, nside_sparse = 16384, bool
>>> map_bool[0] = True
ValueError                                Traceback (most recent call last)
<ipython-input-10-0edf9ce03ec2> in <module>
----> 1 map_bool[0] =  True

/cvmfs/des.opensciencegrid.org/fnal/anaconda2/envs/des20a/lib/python3.7/site-packages/healsparse/healSparseMap.py in __setitem__(self, key, value)
   1421         if isinstance(key, int):
   1422             # Set a single pixel
-> 1423             return self.update_values_pix(np.array([key]), value)
   1424         elif isinstance(key, slice):
   1425             # Set a slice of pixels

/cvmfs/des.opensciencegrid.org/fnal/anaconda2/envs/des20a/lib/python3.7/site-packages/healsparse/healSparseMap.py in update_values_pix(self, pixels, values, nest, operation)
    570                 if isinstance(values, numbers.Integral):
    571                     if not self.is_integer_map:
--> 572                         raise ValueError("Cannot set non-integer map with an integer")
    573                     is_single_value = True
    574                     _values = np.array([values], dtype=self.dtype)

ValueError: Cannot set non-integer map with an integer
```
However, everyone seems happy with `map_bool[0] = float(True)`. Probably need to define `is_boolean_map` or hack `is_integer_map` to be True for boolean maps.

kadrlica:
Looks like the sentinel values for boolean maps are also messed up (they should be `False` instead of `None`):
```
>>> hsp.HealSparseMap.make_empty(32, 16384, np.bool)._sentinel is None           
True
```

erykoff:
Fixed on #129

PR #123, opened 2022-01-29
Add support for boolean-valued maps.

erykoff:
Note that these can only be serialized with parquet and not fits.

### DIFF
--- a/healsparse/fits_shim.py
+++ b/healsparse/fits_shim.py
@@ -132,9 +132,9 @@ class HealSparseFits(object):
         else:
             hdu = self.fits_object[extension]
             if row_range is None:
-                return hdu.data.view(np.ndarray, memmap=False)
+                return hdu.data.view(np.ndarray)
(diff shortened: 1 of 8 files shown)


erykoff:
This addresses #122.

esheldon:
why doesn't it work with fits?

erykoff:
Fits doesn’t support Boolean images. https://heasarc.gsfc.nasa.gov/docs/software/fitsio/c/c_user/node40.html (Also not supported in fitsio). You can use them as a column in a binary table but that is separate in healsparse.

esheldon:
You could use a small integer type.

erykoff:
Okay, I just tried something but can make it better and do it automatically on serialization I think.

fjaviersanchez (approved):
Thanks @erykoff. It looks good to me.

erykoff:
This was properly fixed on #129

PR #127, opened 2022-05-13
Add native support for reading/writing explicit/partial healpix maps.

erykoff:
@gpdf requested that healsparse be able to write out HEALPix EXPLICIT/PARTIAL maps (https://healpix.sourceforge.io/data/examples/healpix_fits_specs.pdf) without going through healpy and requiring the full sky map.  This was fairly easy to implement, and I also added native support for reading EXPLICIT/PARTIAL maps which will save a significant amount of memory/time when reading in one of these maps.

Note that this is not a preferred way of storing healsparse maps because of loss of coverage map info, compression, and fast partial file reads.

### DIFF
--- a/healsparse/io_map_healpix.py
+++ b/healsparse/io_map_healpix.py
@@ -1,0 +1,56 @@
+import os
+import numpy as np
+import warnings
+
+from .fits_shim import _make_header, _write_healpix_filename
+
+
+def _write_map_healpix(hsp_map, filename, clobber=False):
+    """
+    Internal method to write a HealSparseMap with healpix EXPLICIT format.
+
+    Note that the coverage map is not persisted in this format, and only
+    floating point value maps are supported.
+
+    Parameters
+    ----------
+    hsp_map : `HealSparseMap`
+        HealSparseMap to write to a file.
+    filename : `str`
+        Name of file to save
+    clobber : `bool`, optional
+        Clobber existing file?  Default is False.
+
+    Raises
+    ------
+    RuntimeError if file exists and clobber is False.
+    NotImplementedError if persisting anything other than a floating point map.
+    """
+
+    if hsp_map.is_rec_array:
+        raise NotImplementedError("The healpix EXPLICIT output format is not supported for recarray maps.")
+
+    if hsp_map.is_integer_map:
+        warnings.warn("Integer maps will be converted to float on read by healpy.", UserWarning)
+
+    if os.path.isfile(filename) and not clobber:
+        raise RuntimeError("Filename %s exists and clobber is False." % (filename))
+
+    valid_pixels = hsp_map.valid_pixels
+
+    hdr = _make_header(hsp_map.metadata)
+    hdr['PIXTYPE'] = 'HEALPIX'
+    hdr['INDXSCHM'] = 'EXPLICIT'
+    hdr['ORDERING'] = 'NESTED'
+    hdr['NSIDE'] = hsp_map._nside_sparse
+    hdr['OBS_NPIX'] = valid_pixels.size
+    hdr['BAD_DATA'] = hsp_map._sentinel
+    hdr['OBJECT'] = 'PARTIAL'
+    hdr['COORDSYS'] = 'C'
+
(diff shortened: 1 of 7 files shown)


gpdf:
Agreed about the "not preferred"!  This is just for interoperability with legacy community tools.

healsparse/io_map_healpix.py
```diff
@@ -0,0 +1,52 @@
...
+        raise RuntimeError("Filename %s exists and clobber is False." % (filename))
+
+    hdr = _make_header(hsp_map.metadata)
+    hdr['PIXTYPE'] = 'HEALPIX'
+    hdr['INDXSCHM'] = 'EXPLICIT'
+    hdr['ORDERING'] = 'NESTED'
+    hdr['NSIDE'] = hsp_map._nside_sparse
+    hdr['BAD_DATA'] = hsp_map._sentinel
+    hdr['OBJECT'] = 'PARTIAL'
+    hdr['COORDSYS'] = 'E'
```
gpdf: Pretty sure you meant 'C' here: "celestial/equatorial". 'E' is "ecliptic".

healsparse/io_map_healpix.py
```diff
@@ -0,0 +1,52 @@
...
+        raise NotImplementedError("The healpix EXPLICIT output format is not supported for integer maps.")
+
+    if os.path.isfile(filename) and not clobber:
+        raise RuntimeError("Filename %s exists and clobber is False." % (filename))
+
+    hdr = _make_header(hsp_map.metadata)
+    hdr['PIXTYPE'] = 'HEALPIX'
+    hdr['INDXSCHM'] = 'EXPLICIT'
+    hdr['ORDERING'] = 'NESTED'
+    hdr['NSIDE'] = hsp_map._nside_sparse
```
gpdf, line 45: An `OBS_NPIX` header is documented as "required" in the HEALPix-FITS-format PDF referenced in this PR, as the "number of pixels listed in this file".  Of course, this is redundant with information already in the FITS binary table, but I think it should be provided.

healsparse/io_map_healpix.py
```diff
@@ -0,0 +1,52 @@
...
+    Raises
+    ------
+    RuntimeError if file exists and clobber is False.
+    NotImplementedError if persisting anything other than a floating point map.
+    """
+
+    if hsp_map.is_rec_array:
+        raise NotImplementedError("The healpix EXPLICIT output format is not supported for recarray maps.")
+    if hsp_map.is_integer_map:
+        raise NotImplementedError("The healpix EXPLICIT output format is not supported for integer maps.")
```
gpdf: I think this is perhaps being a bit too strict - most likely integer columns would be OK nowadays.  In particular, there is a hint in the [healpy 1.15 release notes](https://github.com/healpy/healpy/releases/tag/1.15.0) that non-float32 datatypes are now supported on I/O.

gpdf: Maybe allow it but print a "Some older tools may not be able to process integer-valued maps" warning?

gpdf (changes requested):
Suggested header changes should be made.  Support for non-float data is just a suggestion and not requested for merge.

healsparse/io_map_healpix.py
erykoff: The release notes refer to the fact that it will now preserve 32-bit or 64-bit floats rather than have to force the user to know what it should be instead of incorrectly converting all the time (which was the annoying default behavior).
You can preserve integer maps in a healpix fits file, but healpy will (still) always convert ints to floats on read and converts all `BAD_DATA` to `hp.UNSEEN`.

But I'll let it save integer maps, which can be read in just fine by healsparse.  But I'll add a warning about healpy.

gpdf: I see; OK.  I didn't look deeply enough.

erykoff: Some of my long-standing frustrations with healpy that we addressed with healsparse.  But it doesn't break, so I've added the feature.

fjaviersanchez (approved):
Changes look good to me. I don't have any comments beyond what @gpdf already mentioned. Thanks a lot!

PR #129, opened 2022-06-08
Add tests for bool maps (and appropriate fixes).

erykoff:
The bool map support was poorly tested and therefore quite buggy.  This PR adds a bunch of tests for boolean maps and fixes the bugs that were revealed.

A lot of the trouble is that `np.bool_` is not related to `bool` in terms of the class hierarchy.


fjaviersanchez (approved):
Changes look good to me. Thanks!

PR #130, opened 2022-06-10
Add support for Multi-Order Coverage (MOC) maps

erykoff:
This PR adds support for reading and writing v1 (NUNIQ) MOC maps: https://www.ivoa.net/documents/MOC/20220317/REC-moc-2.0-20220317.pdf

This may or may not prove useful.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ["3.8", "3.9", "3.10"]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -34,6 +34,7 @@ jobs:
         flake8
     - name: Test with pytest
       run: |
+        pip install mocpy
         pytest
         pip install fitsio>=1.0.5
         pytest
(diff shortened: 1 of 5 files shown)


.github/workflows/python-package.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ["3.8", "3.9", "3.10"]
```
fjaviersanchez: Do we want to drop 3.7?

healsparse/io_map_fits.py
```diff
@@ -511,3 +537,62 @@ def _write_map_fits(hsp_map, filename, clobber=False, nocompress=False):
...
+    order = np.round(np.log2(tbl['UNIQ']//4)).astype(np.int32)//2
+    moc_order = np.max(order)
+
+    hdu = fits.BinTableHDU(tbl)
+    hdu.header['PIXTYPE'] = 'HEALPIX'
+    hdu.header['ORDERING'] = 'NUNIQ'
+    hdu.header['COORDSYS'] = 'C'
+    hdu.header['MOCORDER'] = moc_order
+    hdu.header['MOCTOOL'] = 'healsparse'
+    hdu.header['MOCVERS'] = '1.1'
```
fjaviersanchez, line 596: I guess this choice is due to `pymoc` using v1.1? Do you know when/if they plan to transition to 2.0?

fjaviersanchez (approved):
The PR looks good to me. I left a couple of questions about python version and MOC version, but they don't really change the PR. Thanks!

.github/workflows/python-package.yml
erykoff: Maybe not.  I can re-add it.

healsparse/io_map_fits.py
erykoff: The 2.0 spec was just released (in March), and I have had a hell of a time figuring out the format because the documentation is unclear and the mocpy fits reading is buggy and at the moment I just didn't have the patience.  I wanted to get something in here right now that works, and expand from there if it proves useful.

fjaviersanchez: thanks. It makes sense. I agree to sticking to 1.1 for now, and see how things evolve. I've been looking for their FITS-IO routines, but they don't seem to be easy to find (I found something in rust that I guess they pull now from the MOC rust core package: https://github.com/cds-astro/mocpy/blob/95fe96b371504449cd4590c9d80049db66be4532/src/moc/src/deser/fits/mod.rs)

fjaviersanchez:
Looks good! Please feel free to merge!

healsparse/io_map_fits.py
erykoff: Yup, it's down in rust ... and at that point I gave up looking further.  The present problem is that astropy will output (perfectly legal) FITS with `TFORM1  = 'K       '`, but the mocpy fits reader (at least the one used in the non-deprecated branch) demands `TFORM1  = '1K      '`.

PR #132, opened 2022-07-12
Use hpgeom as healpix geometry backend, replacing healpy.

erykoff:
This PR is a major change, replacing `healpy` as the healpix geometry backend with the much simpler `hpgeom`.  This will lead to faster import times, fewer spurious warnings, and the ability to add new shapes to the geometry library such as ellipses and boxes of constant ra/dec.

See https://hpgeom.readthedocs.io/en/latest/ for details on `hpgeom`.


fjaviersanchez (approved):
It looks good. I also downloaded it to my laptop and works. I searched for stray healpys and didn't find any, so I think we are good to go :)

esheldon:
sorry I missed this, but looks fine

PR #138, opened 2023-03-21
Add ability to set pixels to None, which will set to the sentinel value.

erykoff:
This also updates the MOC support to be compatible with newer mocpy, and makes the `sentinel` getter public.


esheldon (approved):
The parts that are relevant to my issue look OK.

PR #139, opened 2023-03-27
Add publish workflow from git tags.

erykoff:
Now we can push tags and automatically publish to pypi.


fjaviersanchez (approved):
Changes look good! Thank you!

PR #142, opened 2023-06-21
Fix deprecation for astropy tile_size

erykoff:
This also fixes a warning about datatype conversions and adds python=3.11 to the build matrix.

### DIFF
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -58,7 +58,6 @@ class HealSparseMap(object):
         if cov_index_map is not None and cov_map is not None:
             raise RuntimeError('Cannot specify both cov_index_map and cov_map')
         if cov_index_map is not None:
-            import warnings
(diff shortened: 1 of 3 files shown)


fjaviersanchez (approved):
Looks good! Thank you

healsparse/healSparseMap.py
```diff
@@ -58,7 +58,6 @@ def __init__(self, cov_map=None, cov_index_map=None, sparse_map=None, nside_spar
         if cov_index_map is not None and cov_map is not None:
             raise RuntimeError('Cannot specify both cov_index_map and cov_map')
         if cov_index_map is not None:
-            import warnings
             warnings.warn("cov_index_map deprecated", DeprecationWarning)
```
timj, line 61: Consider adding `stacklevel=2` here so that people have an idea what in their code is causing the deprecation warning.

PR #143, opened 2023-07-25
Add tests for empty wide masks, and operations tests.

erykoff:
This PR adds some tests for the type returned when there are no pixels in a wide mask, and fixes for the bugs found when trying to do that,


erykoff:
I also fixed a bug where setting a zero-length array of pixels with a zero-length array of values would raise; that is okay.  At the same time, I changed this raise to a warning because it's possible to just continue with a no-op, but it's probably not something you want to do.

fjaviersanchez (approved):
Changes look good, and seem to work in my computers.

Thank you!

PR #145, opened 2023-08-11
Add bilinear interpolation to healsparse maps.

erykoff:
This PR uses the new hpgeom bilinear interpolation functionality to add interpolation to the sparse maps.

The interpolation routine checks if any of the 4 interpolation neighbors is not valid.  In the default mode, in this case the interpolated value is set to UNSEEN.  In the `allow_partial=True` mode the valid neighbors will be used to approximate the interpolation.

Note that healpy (at least older versions) will just interpolate with the UNSEEN map values, giving garbage that is neither UNSEEN or a useful value.

### DIFF
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -1535,6 +1535,68 @@ class HealSparseMap(object):
 
         return new_map
 
+    def interpolate_pos(self, ra_or_theta, dec_or_phi, lonlat=True, allow_partial=False):
+        """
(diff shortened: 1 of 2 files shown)


erykoff:
I am perfectly happy to have suggestions if the default method name (`interpolate_pos`) or the invalid override (`allow_partial`) are unclear or lacking in any way.

fjaviersanchez (approved):
Looks good! Thank you @erykoff

PR #148, opened 2023-11-04
Fix writing/reading partial parquet files

erykoff:
This fixes row group sizing with parquet writing and therefore fixes reading of partial parquet maps.

It also removes a spurious warning when an empty pixel list is set.


esheldon (approved):

beckermr (approved):
I had not appreciated that you are constructing the partitioned parquet data by hand. This comment isn't for this PR, but there is in principle more than one way to construct the directory names. 

LGTM!

erykoff:
@beckermr I think these are the standard directory names.  The reason to build by hand like this is that it allows (future) parallelization and much saving of memory.

PR #149, opened 2023-11-07
Add bit packed maps to HealSparseMap

erykoff:
This saves 8x memory for boolean masks.

### DIFF
--- a/healsparse/__init__.py
+++ b/healsparse/__init__.py
@@ -24,4 +24,3 @@ from . import geom
 from .geom import Circle, Ellipse, Polygon, realize_geom
 from .utils import WIDE_MASK
 from .cat_healsparse_files import cat_healsparse_files
-from .bitSparseMap import BitSparseMap
--- a/healsparse/__init__.py
+++ b/healsparse/__init__.py
@@ -24,3 +24,4 @@ from . import geom
 from .geom import Circle, Ellipse, Polygon, realize_geom
 from .utils import WIDE_MASK
 from .cat_healsparse_files import cat_healsparse_files
+from .bitarray import BitSparseMap
--- a/healsparse/__init__.py
+++ b/healsparse/__init__.py
@@ -24,4 +24,4 @@ from . import geom
 from .geom import Circle, Ellipse, Polygon, realize_geom
 from .utils import WIDE_MASK
 from .cat_healsparse_files import cat_healsparse_files
-from .bitarray import BitSparseMap
+from .bitSparseMap import BitSparseMap
--- a/healsparse/bitSparseMap.py
+++ b/healsparse/bitSparseMap.py
@@ -13,15 +13,14 @@ class BitSparseMap:
             if not isinstance(data_buffer, np.ndarray) or data_buffer.dtype != np.uint8:
                 raise ValueError("data_buffer must be a numpy array of type uint8")
 
-            self._size = len(data_buffer) * 8
             self._data = data_buffer
         else:
             # Check if size is multiple of 8.
             if (size % 8) != 0:
                 raise ValueError("BitSparseMap must have a size that is a multiple of 8.")
 
-            self._size = size
             self._data = np.zeros(size // 8, dtype=np.uint8)
+            # We need to rething this part.
             if fill_value:
                 # Set to all 1s
                 self._data[:] = 255
@@ -41,22 +40,17 @@ class BitSparseMap:
 
     @property
     def shape(self):
-        return (self._size, )
+        return (self.size, )
 
     @property
     def size(self):
-        return self._size
+        return len(self._data) * 8
 
     def resize(self, newsize, refcheck=False):
         if (newsize % 8) != 0:
             raise ValueError("BitSparseMap must have a size that is a multiple of 8.")
 
         self._data.resize(newsize // 8, refcheck=refcheck)
-        self._size = newsize
-
-    def reshape(self, newshape):
-        # This is a tricky one.
-        pass
 
     def sum(self, shape=None, axis=None):
         if shape is None:
@@ -68,7 +62,7 @@ class BitSparseMap:
                 raise ValueError(f"Axis {axis} is out of bounds for shape.")
             # The shape needs to (a) the last axis must be a multiple
             # of 8; the product needs to equal ths size.
-            if np.prod(shape) != self._size:
+            if np.prod(shape) != self.size:
                 raise ValueError("Shape mismatch with array size.")
             if shape[-1] % 8 != 0:
                 raise ValueError("Final shape index must be a multiple of 8.")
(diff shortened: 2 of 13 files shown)


healsparse/fits_shim.py
```diff
@@ -224,61 +236,111 @@ def _write_filename(filename, c_hdr, s_hdr, cov_index_map, sparse_map,
...
+                    _sparse_map,
+                    extname=s_hdr["EXTNAME"],
+                    header=s_hdr,
+                    compress="GZIP_2",
+                    tile_dims=_tile_dims,
+                    qlevel=0.0,
+                    qmethod=None,
+                )
+            else:
+                f.write(sparse_map, extname=s_hdr["EXTNAME"], header=s_hdr)
```
esheldon, line 283: the docs said that uncompressed also have this 2**31-1 limitation, did you find that is not the case?

healsparse/healSparseMap.py
```diff
@@ -1195,6 +1242,10 @@ def valid_pixels(self):
             valid_pixel_inds, = np.where(self._sparse_map[self._primary] != self._sentinel)
         elif self._is_wide_mask:
             valid_pixel_inds, = np.where(np.any(self._sparse_map != self._sentinel, axis=1))
+        elif self._is_bit_mask:
```
esheldon: Is the way around this to work in coverage pixels?

healsparse/fits_shim.py
erykoff: I have not found this limitation for uncompressed files.

healsparse/healSparseMap.py
erykoff: Yup.  I'm thinking about new APIs for `valid_pixels_chunks` (or something with a better name; ideas welcome) that will give a nice iterator.  I also know how to make the coverage generator a view rather than a copy which will make it easier to use.

esheldon (approved):

healsparse/bitSparseMap.py
```diff
@@ -0,0 +1,310 @@
...
+        pass
+
+    def __ior__(self, other):
+        pass
+
+    def __xor__(self, other):
+        pass
+
+    def __ixor__(self, other):
+        pass
```
beckermr: As far as I can tell, these methods will fail silently if anyone tried to use them. I'd rather they raise a NotImplementedError or we add tests showing they work.

tests/test_bitsparsemap.py
```diff
@@ -0,0 +1,431 @@
...
+        testing.assert_array_equal(fracdet.valid_pixels, fracdet_bool.valid_pixels)
+        testing.assert_array_almost_equal(
+            fracdet[fracdet.valid_pixels],
+            fracdet_bool[fracdet_bool.valid_pixels],
+        )
+
+    def test_bit_mask_from_other_map(self):
+        nside_coverage = 32
+
+        for mode in ['regular', 'wide', 'recarray']:
```
beckermr: mode is not set to `bitmask` in this loop.

beckermr: Also is it bitmask or bit_mask?

tests/test_bitsparsemap.py
```diff
@@ -0,0 +1,431 @@
...
+
+        testing.assert_array_equal(sparse_map_in_nocomp.valid_pixels, sparse_map_in.valid_pixels)
+
+    def test_bit_mask_fracdet(self):
+        nside_coverage = 32
+        nside_map = 2**13
+
+        # We compare the fracdet map generated with the bit_mask code to that
+        # generated with a regular boolean map.
+        sparse_map = HealSparseMap.make_empty(nside_coverage, nside_map, np.bool_, bit_mask=True)
```
beckermr:
```suggestion
        sparse_map = HealSparseMap.make_empty(nside_coverage, nside_map, np.bool_, bit_packed=True)
```

This might be a slightly better name for the API option.

healsparse/__init__.py
```diff
@@ -24,3 +24,4 @@
 from .geom import Circle, Ellipse, Polygon, realize_geom
 from .utils import WIDE_MASK
 from .cat_healsparse_files import cat_healsparse_files
+from .bitSparseMap import BitSparseMap
```
beckermr: This is an internal construct and so I don't think should be part of the public API.

tests/test_packedboolarray.py
```diff
@@ -0,0 +1,432 @@
...
+        # Set True
+        m[0: 64] = True
+        testing.assert_array_equal(m[0: 64], True)
+        testing.assert_array_equal(m[64:], False)
+
+        # Set False
+        m[16: 32] = False
+        testing.assert_array_equal(m[0: 16], True)
+        testing.assert_array_equal(m[16: 32], False)
+        testing.assert_array_equal(m[64:], False)
```
beckermr, line 103:
```suggestion
        testing.assert_array_equal(m[32:], False)
```

tests/test_packedboolarray.py
```diff
@@ -0,0 +1,432 @@
...
+
+        inds = [1, 5, 10, 20]
+        m[inds] = True
+
+        testing.assert_array_equal(m[inds], True)
+
+        values = np.array([True, False, True, True])
+        m[tuple(inds)] = values
+
+        testing.assert_array_equal(m[tuple(inds)], values)
```
beckermr, line 165: Why does one need the tuple here?

tests/test_packedboolarray.py
```diff
@@ -0,0 +1,432 @@
...
+        m2 = m[16: 64]
+        m2[0] = False
+        self.assertEqual(m[16], False)
+
+        with self.assertRaises(ValueError):
+            _ = m[5: 32]
+        with self.assertRaises(ValueError):
+            _ = m[8: 31]
+        with self.assertRaises(ValueError):
+            _ = m[8: 32: 4]
```
beckermr, line 190: These raise because they are not multiples of 8?

tests/test_packedboolarray.py
erykoff: Changed to more complete (and correct):
```
        testing.assert_array_equal(m[0: 16], True)
        testing.assert_array_equal(m[16: 32], False)
        testing.assert_array_equal(m[32: 64], True)
        testing.assert_array_equal(m[64:], False)
```

tests/test_packedboolarray.py
erykoff: I wanted to test both list and tuple.

tests/test_packedboolarray.py
beckermr: ahhh woops! kudos!

beckermr (approved):
Needs docs but otherwise LGTM!

beckermr (approved):

PR #150, opened 2023-11-15
Add boolean map operations and other optimizations.

erykoff:
This PR adds operations for boolean maps (both numpy array and bit packed) to make it easy to do `&`, `|`, `^`, and `~`.  This makes it straightforward to combine maps.

This also adds a convenience routine to get valid pixels per coverage pixel to save memory when using very large masks.

### DIFF
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -1920,7 +1920,6 @@ class HealSparseMap(object):
 
         Cannot be used with recarray maps.
         """
-
         return self._apply_operation(other, np.add)
 
     def __iadd__(self, other):
@@ -2010,8 +2009,10 @@ class HealSparseMap(object):
 
         Cannot be used with recarray maps.
         """
-
-        return self._apply_operation(other, np.bitwise_and, int_only=True)
+        if self.dtype == np.bool_:
+            return self._apply_boolean_map_operation(other, "and")
+        else:
+            return self._apply_operation(other, np.bitwise_and, int_only=True)
 
     def __iand__(self, other):
         """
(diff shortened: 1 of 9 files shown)


healsparse/healSparseMap.py
```diff
@@ -1313,6 +1313,58 @@ def n_valid(self):
         self._n_valid = n_valid
         return n_valid
 
+    def get_valid_pixels_per_covpix(self):
+        """
+        Get all single covpixel valid_pixels, one at a time.
```
esheldon: This makes it sound like you return on pixel at a time.  Maybe "Generator to get valid_pixels associate with each coverage pixel, yielded one coverage pixel at a time"

Then it would be good to show usage since this is a generator.

esheldon: Also I don't see a test for this

esheldon (changes requested):

healsparse/healSparseMap.py
erykoff: One of my big questions was about naming this routine.

tests/test_single_covpix.py
```diff
@@ -119,11 +141,17 @@ def test_single_pixel_generator(self):
...
         for i, sub_map in enumerate(sparse_map.get_covpix_maps()):
             sub_cov_pixels, = np.where(sub_map.coverage_mask)
+
             self.assertEqual(len(sub_cov_pixels), 1)
             self.assertEqual(sub_cov_pixels[0], cov_pixels[i])
             self.assertEqual(sub_map.n_valid, 1)
+            testing.assert_array_equal(sub_map.valid_pixels, pixels[i])
             testing.assert_almost_equal(sub_map[pixels[i]], 1.0)
 
+        for i, sub_valid_pixels in enumerate(sparse_map.get_valid_pixels_per_covpix()):
```
erykoff: @esheldon This is the test for the new generator.

healsparse/healSparseMap.py
esheldon: Other ideas:
```
iter_valid_pixels
iter_valid_pixels_by_covpix
```

esheldon:
Do you expect to add more or should I finalize my review?

erykoff:
This is it for this PR unless you think I missed something.

esheldon (approved):

PR #151, opened 2024-01-20
Add support for pixel ranges and composable geometries.

erykoff:
This completely refactors the packed boolean arrays to be more efficient, and to work with arbitrary slices.  This enables setting of giant pixel ranges (large area masks) with minimal memory overhead.


tests/test_pixel_ranges.py
```diff
@@ -0,0 +1,237 @@
...
+import healsparse
+
+
+class UpdateValuesPixelRangesTestCase(unittest.TestCase):
+    def test_update_values_pixel_ranges(self):
+        nside_coverage = 32
+        nside_map = 256
+
+        # Monkey-patch PIXEL_RANGE_THRESHOLD for testing.
+        healsparse.healSparseMap.PIXEL_RANGE_THRESHOLD = 0
```
beckermr, line 16: Don't you need to setup and teardown this patching?

beckermr: Same for the ones below.

healsparse/packedBoolArray.py
```diff
@@ -7,33 +7,66 @@
...
     """
-    def __init__(self, size=0, fill_value=False, data_buffer=None):
+    def __init__(self, size=None, data_buffer=None, start_index=None, stop_index=None):
+        if size is not None and data_buffer is not None:
+            raise ValueError("May only specify one of size or data_buffer.")
+
+        if start_index is None:
+            self._start_index = 0
+        else:
+            if start_index < 0 or start_index > 7:
```
beckermr, line 37: I don't understand why this is a constraint. Sure the user could waste memory, but do we care if they do something less optimal?

healsparse/packedBoolArray.py
```diff
@@ -81,18 +123,52 @@ def shape(self):
 
     @property
     def size(self):
-        return len(self._data) * 8
+        return self._stop_index - self._start_index
```
beckermr, line 128: Don't you need a `+1` here since if you have, eg., start_index = 0 and stop_index = 9, there are 10 elements?

beckermr: `stop_index` is defined as the "index at which the array ends" in the doc string. However, given how it is computed and used, I think it is actual one plus the last index.

healsparse/packedBoolArray.py
```diff
@@ -111,6 +189,9 @@ def sum(self, shape=None, axis=None):
 
     @property
     def data_array(self):
+        if self._start_index != 0 or self._stop_index != self.size:
+            raise NotImplementedError("Cannot yet get data array from offset array")
```
beckermr: This error message confuses me a lot.

healsparse/packedBoolArray.py
```diff
@@ -120,164 +201,181 @@ def copy(self):
         -------
         copy : `_PackedBoolArray`
         """
-        return _PackedBoolArray(data_buffer=self._data.copy())
+        first_unpacked, mid_data, last_unpacked = self._extract_first_middle_last(mask_extra=True)
+        new_buffer = self._data.copy()
+        if first_unpacked[0] is not None:
+            new_buffer[0] = np.packbits(first_unpacked[0], bitorder="little")
+        if last_unpacked[0] is not None:
+            new_buffer[-1] = np.packbits(last_unpacked[0], bitorder="little")
```
beckermr, line 216: Why is this block needed? Since start and stop don't change, isn't the existing data already packaged correctly? Maybe add a comment?

healsparse/packedBoolArray.py
```diff
@@ -288,67 +386,259 @@ def __setitem__(self, key, value):
...
+                    if len(mid_data) == 0:
+                        mid_data = None
+                    return (
+                        (first_unpacked, self._start_index, None),
+                        mid_data,
+                        (last_unpacked, 0, self._stop_index % 8),
+                    )
+
+        # Should not get here.
+        raise RuntimeError("Programmer mistake")
```
beckermr: This needs a better error message.

tests/test_packedboolarray.py
```diff
@@ -13,378 +13,1094 @@
...
+        array = np.zeros(16, dtype=np.bool_)
+        array[0] = True
+        array[4] = True
+        array[7] = True
+        array[12] = True
+        array[15] = True
+
+        data = np.packbits(array, bitorder="little")
+
+        middle_arrays.append(_PackedBoolArray(data_buffer=data.copy(), start_index=0, stop_index=16))
```
beckermr, line 57: The last index is 15 in this case. Another example of how the doc string and parameter names do not match how they are used.

tests/test_packedboolarray.py
```diff
@@ -13,378 +13,1094 @@
...
-            arr = np.zeros(128, dtype=np.int32)
-            m = _PackedBoolArray.from_boolean_array(arr)
-
-    def test_resize(self):
-        m = _PackedBoolArray(size=0)
-        m.resize(2**10)
-        self.assertEqual(m.size, 2**10)
-        self.assertEqual(len(m.data_array), 2**10 // 8)
-        testing.assert_array_equal(m.data_array, 0)
+            pba = _PackedBoolArray(size=20, start_index=-1)
```
beckermr, line 115: -1? That is allowed?

beckermr (changes requested):
I left a few comments. Congrats on the test suite. It was painful to read and no doubt it was painful to write!

tests/test_packedboolarray.py
beckermr: woops missed the assertRaises in the confusing diff! sorry!

healsparse/packedBoolArray.py
erykoff: The unaligned ends of the array need special handling because they need to be unpacked to be able to use them.  And if I allow another trailing bit of data that shouldn't be touched when doing an operation then I think I have to make that already annoying logic even more complex.

healsparse/packedBoolArray.py
erykoff: Ah yes, it is the exclusive stop index, like a python slice.  I will correct the docstring.

healsparse/packedBoolArray.py
erykoff: Though now that I think about it some more, I can always just create a view on construction.  But after spending 15 minutes on this, I have broken everything terribly and don't have time to redo all the index math for the fully general case.

healsparse/packedBoolArray.py
erykoff: How about `_PackedBoolArray does not support extracting the data_array from an unaligned _PackedBoolArray.`. This is one of those "it kind of works but will give surprising results in some cases".

healsparse/packedBoolArray.py
erykoff: I want to mask any extraneous bits which may have any value depending on how things were originally sliced.

healsparse/packedBoolArray.py
erykoff: Should just be removed.

tests/test_packedboolarray.py
erykoff: Fixed the docstring.

healsparse/packedBoolArray.py
beckermr: understood!

beckermr (approved):

Issue #154, opened 2024-03-14
print of map uses huge memory


esheldon:
This uses more memory than 12G (I don't know how much because it was killed).  Maybe in a case like this it is better to just print some summary information about the map

```
[esheldon@forest y6-healsparse] ipython
[nav] In [1]: import healsparse
[nav] In [2]: hmap = healsparse.HealSparseMap.read('y6-combined-hleda-gaiafull-des-stars-hsmap131k-mdet-v2.hsp')
[ins] In [3]: hmap
```

erykoff:
Okay, can replicate locally.  I will investigate.

PR #155, opened 2024-03-15
Quick fix for high memory usage when printing summary of a bit packed map.

erykoff:
As pointed out in #154 printing a large bit packed map can use a lot of memory.  To truly fix this requires some c code, which will be added in the near future.  In the meantime, we can fix the problem by not auto-printing the number of pixels.

This also uses a slightly more memory efficient lookup table for bit counting.


esheldon:
print is fast and I checked LUT is working

esheldon (approved):

PR #157, opened 2024-04-17
Fix crash when reading a pathlib.Path type object.

### LINKED ISSUE #156: healsparse won't read a `pathlib.Path`
When trying to read from a `pathlib.Path` object, healsparse will throw an error about only supporting fits and parquet files; e.g.,
```
(Pdb) hsp_plus
PosixPath('/pscratch/sd/s/smau/y6-image-sims/debug-cosmos-rand-piff-median_color-seed/DES0608-5414/16200/plus/des-pizza-slices-y6/DES0608-5414/metadetect/DES0608-5414_metadetect-config_mdetcat_part0000-healsparse-mask.hs')
(Pdb) healsparse.HealSparseMap.read(hsp_plus)
*** NotImplementedError: HealSparse only supports fits and parquet files.
(Pdb) healsparse.HealSparseMap.read(hsp_plus.as_posix())
HealSparseMap: nside_coverage = 128, nside_sparse = 131072, int32, 2551130 valid pixels
```
the error is also slightly confusing as the file format is not a problem.

This is happening because a `Path` doesn't match as a `str` (https://github.com/LSSTDESC/healsparse/blob/833509256873ae92db52af8e15e17ff9d9559a01/healsparse/io_coverage.py#L33-L46)

One possible solution would be checking 
```
isinstance(filename_or_fits, (str, Path))
```
instead of 
```
isinstance(filename_or_fits, str)
```

I haven't checked if any of the code downstream would complain about a `Path`...

erykoff:
This fixes #156 .

### DIFF
--- a/healsparse/io_coverage.py
+++ b/healsparse/io_coverage.py
@@ -1,4 +1,5 @@
 import os
+import pathlib
 
 from .fits_shim import HealSparseFits
 from .io_coverage_fits import _read_coverage_fits
@@ -29,6 +30,9 @@ def _read_coverage(coverage_class, filename_or_fits, use_threads=False):
     """
     is_fits = False
     is_parquet_file = False
+
+    if isinstance(filename_or_fits, pathlib.PurePath):
+        filename_or_fits = os.fspath(filename_or_fits)
 
     if isinstance(filename_or_fits, str):
         try:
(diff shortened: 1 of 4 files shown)


fjaviersanchez (approved):
Looks good to me!

PR #159, opened 2024-04-26
Add nside_render to geom objects to allow for rendering at a different (coarser) resolution than the corresponding map.

erykoff:
This also fixes a bug when trying to use `get_map()` with a boolean map, and adds a `Box` geometry type with sides of constant ra/dec.

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -4,7 +4,9 @@
 HealSparse Geometry
 ===================
 
-:code:`HealSparse` has a basic geometry library that allows you to generate maps from circles, convex polygons, and ellipses as supported by :code:`hpgeom`.  Each geometric object is associated with a single value.  On construction, geometry objects only contain information about the shape, and they are only rendered onto a `HEALPix` grid when requested.
+:code:`HealSparse` has a basic geometry library that allows you to generate maps from circles, convex polygons, ellipses, and boxes as supported by :code:`hpgeom`.
+Each geometric object is associated with a single value.
+On construction, geometry objects only contain information about the shape, and they are only rendered onto a `HEALPix` grid when requested.
 
 There are a few methods to realize geometry objects.
 The easiest is to combine a geometric object with a :code:`HealSparseMap` map, with the ``or``, ``and``, or ``add`` operation.
@@ -15,7 +17,8 @@ Finally, for integer-value objects one can use the :code:`realize_geom()` method
 HealSparse Geometry Shapes
 --------------------------
 
-The three shapes supported are :code:`Circle`, :code:`Ellipse`, and :code:`Polygon`.  They share a base class, and while the instantiation is different, the operations are the same.
+The four shapes supported are :code:`Circle`, :code:`Ellipse`, :code:`Polygon`, and :code:`Box`.
+They share a base class, and while the instantiation is different, the operations are the same.
 
 **Circle**
 
@@ -49,11 +52,23 @@ The three shapes supported are :code:`Circle`, :code:`Ellipse`, and :code:`Polyg
                               value=8)
 
 
+**Box**
+
+.. code-block :: python
+
+    # All units are decimal degrees
+    # A box differs from a polygon in that the sides will be constant ra/dec rather
+    # than great circles.
+    # See https://hpgeom.readthedocs.io/en/latest .
+    box = healsparse.Box(ra1=20.0, ra2=30.0, dec1=10.0, dec2=5.0, value=True)
+
+
 Combining Geometric Objects with Maps
 -------------------------------------
 
 Given a map, it is very simple to combine geometric objects to build up complex shapes/masks/etc.
 Behind the scenes, large geometric objects are rendered with :code:`hpgeom` pixel ranges which leads to greater memory efficiency.
+Note that these operations can be applied to integer or boolean maps.
 
 .. code-block :: python
 
@@ -89,7 +104,8 @@ To make a map from a geometry object, use the :code:`get_map()` method as such. 
 Using :code:`realize_geom()`
 ----------------------------
 
-You can only use :code:`realize_geom()` to create maps from combinations of polygons if you are using integer maps, and want to :code:`or` them together.  This method is more memory efficient than generating each individual individual map and combining them, as above.
+You can only use :code:`realize_geom()` to create maps from combinations of polygons if you are using integer maps, and want to :code:`or` them together.
+This method is more memory efficient than generating each individual individual map and combining them, as above.
 
 .. code-block :: python
 
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -8,6 +8,11 @@ HealSparse Geometry
 Each geometric object is associated with a single value.
 On construction, geometry objects only contain information about the shape, and they are only rendered onto a `HEALPix` grid when requested.
 
+In addition to information about the shape itself, a geometric object may optionally contain a value of :code:`nside_render`.
+This indicates that the shape should always be rendered at this given resolution, no matter the resolution of the map that it is being combined with.
+(Note that you can only render at a resolution that is less than or equal to the map resolution, or else a :code:`ValueError` is raised.)
+This functionality may be useful if one is building a map that may be used with multiple resolutions, and one wants to ensure that a higher and lower resolution maps have exactly the same outline for these shapes.
+
 There are a few methods to realize geometry objects.
 The easiest is to combine a geometric object with a :code:`HealSparseMap` map, with the ``or``, ``and``, or ``add`` operation.
 One can generate a :code:`HealSparseMap` from the geometric object.
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -12,6 +12,7 @@ In addition to information about the shape itself, a geometric object may option
 This indicates that the shape should always be rendered at this given resolution, no matter the resolution of the map that it is being combined with.
 (Note that you can only render at a resolution that is less than or equal to the map resolution, or else a :code:`ValueError` is raised.)
 This functionality may be useful if one is building a map that may be used with multiple resolutions, and one wants to ensure that a higher and lower resolution maps have exactly the same outline for these shapes.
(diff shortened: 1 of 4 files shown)


docs/geometry.rst
```diff
@@ -49,11 +58,23 @@ The three shapes supported are :code:`Circle`, :code:`Ellipse`, and :code:`Polyg
...
+    # See https://hpgeom.readthedocs.io/en/latest .
+    box = healsparse.Box(ra1=20.0, ra2=30.0, dec1=10.0, dec2=5.0, value=True)
+
+
 Combining Geometric Objects with Maps
 -------------------------------------
 
 Given a map, it is very simple to combine geometric objects to build up complex shapes/masks/etc.
 Behind the scenes, large geometric objects are rendered with :code:`hpgeom` pixel ranges which leads to greater memory efficiency.
+Note that these operations can be applied to integer or boolean maps.
```
beckermr, line 77: What happens if one mixes a boolean `value` for the geometry but applies it to an integer map or vice versa? Does healsparse do casts so that 0 is equivalent False and non-zero is True and True casts to 1?

This part of the API confuses me.

healsparse/geom.py
```diff
@@ -233,18 +250,37 @@ def radius(self):
...
+        pixels = hpg.query_circle(
+            nside_render,
             self._ra,
             self._dec,
             self._radius,
             nest=True,
             inclusive=False,
         )
 
+        if self._nside_render is not None:
```
beckermr: The subtle addition of `self.` here is what short circuts this call when nside_render is not set. Maybe rename the `nside_render` variable above to make things clearer?

healsparse/geom.py
```diff
@@ -233,18 +250,37 @@ def radius(self):
...
-            nside,
+        if self._nside_render is not None:
+            if nside < self._nside_render:
+                raise ValueError(f"Cannot render a Circle with {self._nside_render} into nside={nside}")
+            nside_render = self._nside_render
+        else:
+            nside_render = nside
+
+        pixels = hpg.query_circle(
+            nside_render,
```
beckermr:
```suggestion
        if self._nside_render is not None:
            if nside < self._nside_render:
                raise ValueError(f"Cannot render a Circle with {self._nside_render} into nside={nside}")
            _nside = self._nside_render
        else:
            _nside = nside

        pixels = hpg.query_circle(
            _nside,
```

healsparse/geom.py
beckermr: See my suggested change.

healsparse/geom.py
```diff
@@ -233,18 +250,37 @@ def radius(self):
...
-            nside,
+        if self._nside_render is not None:
+            if nside < self._nside_render:
+                raise ValueError(f"Cannot render a Circle with {self._nside_render} into nside={nside}")
+            nside_render = self._nside_render
+        else:
+            nside_render = nside
+
+        pixel_ranges = hpg.query_circle(
+            nside_render,
```
beckermr:
```suggestion
        if self._nside_render is not None:
            if nside < self._nside_render:
                raise ValueError(f"Cannot render a Circle with {self._nside_render} into nside={nside}")
            _nside = self._nside_render
        else:
            _nside = nside

        pixel_ranges = hpg.query_circle(
            _nside,
```

healsparse/geom.py
```diff
@@ -233,18 +250,37 @@ def radius(self):
...
             self._ra,
             self._dec,
             self._radius,
             nest=True,
             inclusive=False,
         )
 
+        if self._nside_render is not None:
+            return hpg.upgrade_pixels(nside_render, pixels, nside)
```
beckermr:
```suggestion
            return hpg.upgrade_pixels(_nside, pixels, nside)
```

healsparse/geom.py
```diff
@@ -253,9 +289,14 @@ def get_pixel_ranges(self, *, nside):
             return_pixel_ranges=True,
         )
 
+        if self._nside_render is not None:
+            return hpg.upgrade_pixel_ranges(nside_render, pixel_ranges, nside)
```
beckermr:
```suggestion
            return hpg.upgrade_pixel_ranges(_nside, pixel_ranges, nside)
```

healsparse/geom.py
```diff
@@ -310,30 +352,54 @@ def vertices(self):
...
             self._dec,
             nest=True,
             inclusive=False,
         )
 
+        if self._nside_render is not None:
+            return hpg.upgrade_pixels(nside_render, pixels, nside)
+        else:
+            return pixels
+
```
beckermr:
```suggestion
        if self._nside_render is not None:
            if nside < self._nside_render:
                raise ValueError(f"Cannot render a Polygon with {self._nside_render} into nside={nside}")
            _nside = self._nside_render
        else:
            _nside = nside

        pixels = hpg.query_polygon(
            _nside,
            self._ra,
            self._dec,
            nest=True,
            inclusive=False,
        )

        if self._nside_render is not None:
            return hpg.upgrade_pixels(_nside, pixels, nside)
        else:
            return pixels

```

healsparse/geom.py
```diff
@@ -310,30 +352,54 @@ def vertices(self):
...
+            nside_render,
             self._ra,
             self._dec,
             nest=True,
             inclusive=False,
             return_pixel_ranges=True,
         )
 
+        if self._nside_render is not None:
+            return hpg.upgrade_pixel_ranges(nside_render, pixel_ranges, nside)
```
beckermr:
```suggestion
        if self._nside_render is not None:
            if nside < self._nside_render:
                raise ValueError(f"Cannot render a Polygon with {self._nside_render} into nside={nside}")
            _nside = self._nside_render
        else:
            _nside = nside

        pixel_ranges = hpg.query_polygon(
            _nside,
            self._ra,
            self._dec,
            nest=True,
            inclusive=False,
            return_pixel_ranges=True,
        )

        if self._nside_render is not None:
            return hpg.upgrade_pixel_ranges(_nside, pixel_ranges, nside)
```

beckermr (changes requested):
One comment on code clarity. The algorithms look correct. There is now some repetition in the logic around handling nside render. One could refactor the classes to defer only the rendering calls to subclasses and lift the handling of nside render into the parent class.

beckermr:
Note I did not flag all of the potential changes from `nside_render` to `_nside`.

docs/geometry.rst
erykoff: Oh this is an interesting one.  The design of healsparse is that it doesn't want to do any implicit conversion.  Certainly it doesn't allow it between numpy types -- you have to explicitly cast.  The problem is that if you use a python type it does allow implicit conversions.  I might need to fix that (would like to hear your thoughts).  For example, if you have an integer map `np.int64` and you try to add in a circle with `value=True` then it'll promote that to `1` (default numpy/python promotion).  But if you do `value=np.bool_(True)` then it raises because of the type mismatch (which is checked when you have two numpy dtypes).

beckermr: OK. What about bit packed boolean arrays? That case is ambiguous to me since I cannot specify an explicitly bit-packed boolean value for the geometry. 

Or if you want, the bit packed bool value for the geometry is 0 and 1. ;p

erykoff: They work the same way.  I know you can't specify a bit packed map for `get_map` but you really should never need `get_map` for anything.  You generate a bit packed or whatever map and then just `|=` or `&=` your shapes into it with a `True` or `False` value.

erykoff:
I did the refactoring, it makes more sense this way.

healsparse/geom.py
```diff
@@ -126,6 +176,8 @@ def get_map(self, *, nside_coverage, nside_sparse, dtype, wide_mask_maxbits=None
         x = np.zeros(1, dtype=dtype)
         if is_integer_value(x[0]):
             sentinel = 0
+        elif dtype == np.bool_:
+            sentinel = False
```
beckermr: What about this?

```
In [1]: import numpy as np

In [2]: np.bool_ == bool
Out[2]: False

```

docs/geometry.rst
beckermr: Well I guess it'd be good to ensure that the code properly errors on non-numpy 0s, 1s, ints, bools, etc. so that it bans implicit conversion completely. We can do this and update the docs in a new PR ofc.

beckermr: This snippet might help:

```
np.array(True).dtype == np.bool_
```

It could be easiest to just convert to numpy right at the start?

beckermr:
```
In [2]: import numpy as np

In [3]: np.array(True).dtype == np.bool_
Out[3]: True

In [4]: np.array(0).dtype
Out[4]: dtype('int64')

In [5]: np.array(1).dtype
Out[5]: dtype('int64')
```

beckermr: #arrayscalarsforthewin?

erykoff: One of the things that I want to preserve is allowing the use of (e.g.) a python int to set an int32 map.  But I need to add range checking.  So this will be more work than just in geom to make sure this works right.

healsparse/geom.py
```diff
@@ -126,6 +176,8 @@ def get_map(self, *, nside_coverage, nside_sparse, dtype, wide_mask_maxbits=None
         x = np.zeros(1, dtype=dtype)
         if is_integer_value(x[0]):
             sentinel = 0
+        elif dtype == np.bool_:
```
beckermr:
```suggestion
        elif dtype == np.bool_ or dtype == bool:
```

beckermr: With this change, I think we can merge this PR.

docs/geometry.rst
beckermr: Sounds good. Let's punt on this issue then. I made on suggestion below to keep the code functional for specifying `dtype=bool` as opposed to `dtype=np.bool_` in `get_map`. With that change, we should be good to go.

beckermr (approved):
LGTM and thank you!

PR #160, opened 2024-05-03
BUG degrees mixed with radians in random draws

beckermr:
It appears that degrees is being mixed with radians in the random draws function? 

I think this has been a silent bug since numerically degrees are always larger. 

However, I wonder if this increases efficiency?

Also, now that this unit has changed, there appears to be a `sin(theta)` factor missing for the draws in `phi`.

And there is no reason to multiply and divide by 1.0.

### DIFF
--- a/healsparse/healSparseRandoms.py
+++ b/healsparse/healSparseRandoms.py
@@ -91,9 +91,10 @@ def make_uniform_randoms(sparse_map, n_random, rng=None):
         sparse_map.nside_coverage,
         units="radians",
     )
+    sintheta = np.sin(cov_theta)
 
-    ra_range = np.clip([np.min(cov_phi - extra_boundary),
-                        np.max(cov_phi + extra_boundary)],
+    ra_range = np.clip([np.min(cov_phi - extra_boundary / sintheta),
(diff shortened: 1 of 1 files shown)


healsparse/healSparseRandoms.py
```diff
@@ -87,10 +87,14 @@ def make_uniform_randoms(sparse_map, n_random, rng=None):
...
-                        np.max(cov_phi + extra_boundary)],
+    extra_boundary = 2.0 * hpg.nside_to_resolution(
+        sparse_map.nside_coverage,
+        units="radians",
+    )
+    sintheta = np.sin(cov_theta)
+
+    ra_range = np.clip([np.min(cov_phi - extra_boundary / sintheta),
+                        np.max(cov_phi + extra_boundary / sintheta)],
                        0.0, 2.0 * np.pi)
```
beckermr:
```suggestion
    ra_range = np.array([
        cov_phi - extra_boundary / sintheta,
        cov_phi + extra_boundary / sintheta,
    ])
```

erykoff (approved):

PR #167, opened 2025-03-03
Fix crash when certain large geometries would cause coverage map to go out-of-bounds.

erykoff:
This PR also updates numpy 2 deprecations and fixes test comparisons that were inconsistent for float32 types for numpy1 vs numpy2.


fjaviersanchez (approved):
Changes look good. Thank you for the updates!

PR #169, opened 2026-01-20
Add support for hdf5 files

elvinpoole:
Added read/write to a group in an hdf5 file

Useful for saving maps in TXPipe/ceci where we want N maps with different masks saved to the same file (e.g. galaxy density maps in different tomo bins)

Currently it only allows the whole map to the read (cannot specify list of coverage pixels with `pixels`) but this can be added

### DIFF
--- a/docs/filespec.rst
+++ b/docs/filespec.rst
@@ -285,7 +285,7 @@ HealSparseMap HDF5 Serialization
 A :code:`HealSparseMap` may also be serialized to an HDF5 file. 
 Multiple :code:`HealSparseMap` objects can be stored in the same HDF5 file.
 Each :code:`HealSparseMap` is stored in a different HDF5 group.
-They are not required to have the same mask, :code:`nside_coverage` or :code:`nside_sparse`.
+They are not required to have the same mask, :code:`nside_coverage`, or :code:`nside_sparse`.
 
 All datasets are written with gzip compression and chunked such that each chunk corresponds to a single coverage pixel.
 
--- a/docs/filespec.rst
+++ b/docs/filespec.rst
@@ -276,3 +276,123 @@ The other columns in the overflow coverage pixel are filled with the default sen
 
 If the sparse map is a bit-packed mask, the schema is the same as for a regular sparse map image.
 In this case, as with the FITS serialization, the sparse map is stored as an array of unsigned 8-bit integers which is the in-memory backing of the bit-packed array.
+
+.. _hdf5_format:
+
+HealSparseMap HDF5 Serialization
+================================
+
+A :code:`HealSparseMap` may also be serialized to an HDF5 file. 
+Multiple :code:`HealSparseMap` objects can be stored in the same HDF5 file.
+Each :code:`HealSparseMap` is stored in a different HDF5 group.
+They are not required to have the same mask, :code:`nside_coverage` or :code:`nside_sparse`.
+
+All datasets are written with gzip compression and chunked such that each chunk corresponds to a single coverage pixel.
+
+HDF5 Group Layout
+-----------------
+
+A serialized :code:`HealSparseMap` is stored within a single HDF5 group (default name :code:`"map"`), which contains:
+
+* A dataset encoding the coverage map
+* One or more datasets encoding the sparse map
+* Attributes storing map metadata
+
+Coverage Map
+------------
+
+The coverage map is stored as a one-dimensional dataset:
+
+* **Dataset name:** :code:`cov_index_map`
+* **Datatype:** :code:`numpy.int64`
+* **Shape:** :code:`(12 * nside_coverage * nside_coverage,)`
+
+This dataset is a direct serialization of the in-memory coverage index map,
+following the same conventions described in :ref:`coverage_map`.
+
+Sparse Map
+----------
+
+The sparse map is stored as a two-dimensional dataset, where each row corresponds to a single coverage pixel.
+The first row always represents the overflow (sentinel) coverage pixel.
+
+Regular Sparse Map
+^^^^^^^^^^^^^^^^^^
+
+For regular (non-record, non-wide-mask, non-bit-packed) sparse maps:
+
+* **Dataset name:** :code:`sparse_map`
+* **Shape:** :code:`(ncov_in_sparse, nfine_per_cov)`
+* **Datatype:** Sparse map datatype
+
+where:
+
+* :code:`ncov_in_sparse`: The number of coverage pixels containing non-sentinel data, plus one for the overflow pixel
+ 
+Each row contains the :code:`nfine_per_cov` sparse pixel values associated with a single coverage pixel.
+The dataset is chunked as :code:`(1, nfine_per_cov)` to allow efficient access to individual coverage pixels.
+
+Sparse Map Record Array
+^^^^^^^^^^^^^^^^^^^^^^^
+
+If the sparse map is a numpy record array, each field is stored in a separate subgroup:
+
+* **Group name:** Name of the record field
+* **Dataset name:** :code:`sparse_map`
+* **Shape:** :code:`(ncov_in_sparse, nfine_per_cov)`
+* **Datatype:** Field datatype
+
+All fields share the same coverage map indexing.
+
+Sparse Map Wide Mask
+^^^^^^^^^^^^^^^^^^^^
+
+For wide mask maps, the sparse map is stored as a three-dimensional dataset:
+
+* **Dataset name:** :code:`sparse_map`
+* **Shape:** :code:`(ncov_in_sparse, nfine_per_cov, wide_mask_width)`
+* **Datatype:** :code:`numpy.uint8`
+
+Each sparse pixel is represented by :code:`wide_mask_width` bytes.
+The dataset is chunked as :code:`(1, nfine_per_cov, wide_mask_width)`.
+
+Sparse Map Bit-Packed Mask
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For bit-packed mask maps, the sparse map is stored in its packed representation:
+
+* **Dataset name:** :code:`sparse_map`
+* **Shape:** :code:`(ncov_in_sparse, nfine_per_cov // 8)`
+* **Datatype:** :code:`numpy.uint8`
+
+Each byte encodes eight sparse pixels.
+
+Metadata and Attributes
+-----------------------
+
+All remaining map metadata is stored as HDF5 group attributes.
+The following attributes are required:
+
+* :code:`nside_sparse`
+* :code:`nside_coverage`
+* :code:`sentinel`
+* :code:`primary`
+* :code:`nest` (always :code:`True`)
+
+The following flags describe the sparse map type:
+
+* :code:`is_rec_array`
+* :code:`is_bit_packed`
+* :code:`is_wide_mask`
+* :code:`wide_mask_width`
+
+Any additional metadata attached to the :code:`HealSparseMap` object is also written as group attributes.
+
+Partial Reads
+-------------
+
+The HDF5 format supports reading a subset of coverage pixels.
+When a subset is requested, only the corresponding rows of the sparse map datasets are read.
+The overflow (sentinel) coverage pixel is always included as the first row.
+
+On read, the sparse map rows are reshaped and reordered to reconstruct the in-memory sparse map layout described in :ref:`sparse_map`.
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -386,7 +386,7 @@ class HealSparseMap(object):
             the ``healpix`` EXPLICIT format does not maintain all metadata and
             coverage information.
         hdf5_group: `str`, optional
-            If file format is ``hdf5``, save to this group, otheriwse unused.
+            If file format is ``hdf5``, save to this group, otherwise unused.
 
         Raises
         ------
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -119,7 +119,7 @@ class HealSparseMap(object):
     @classmethod
     def read(cls, filename, nside_coverage=None, pixels=None, header=False,
              degrade_nside=None, weightfile=None, reduction='mean',
-             use_threads=False, group='map'):
+             use_threads=False, hdf5_group='map'):
         """
         Read in a HealSparseMap.
 
@@ -149,9 +149,8 @@ class HealSparseMap(object):
            Not yet implemented for parquet files.
         use_threads : `bool`, optional
            Use multithreaded reading for parquet files.
-        group : `str`, optional
-            if file is hdf5, read from this group
-            default is `map`
+        hdf5_group : `str`, optional
+           If file format is ``hdf5``, read from this group, otheriwse unused.
 
         Returns
         -------
@@ -163,7 +162,7 @@ class HealSparseMap(object):
         return _read_map(cls, filename, nside_coverage=nside_coverage, pixels=pixels,
                          header=header, degrade_nside=degrade_nside,
                          weightfile=weightfile, reduction=reduction, use_threads=use_threads,
-                         hdf5_group=group)
+                         hdf5_group=hdf5_group)
 
     @classmethod
     def make_empty(cls, nside_coverage, nside_sparse, dtype, primary=None, sentinel=None,
@@ -363,7 +362,7 @@ class HealSparseMap(object):
 
         return cov_map, sparse_map
 
-    def write(self, filename, clobber=False, nocompress=False, format='fits', nside_io=4, group='map'):
+    def write(self, filename, clobber=False, nocompress=False, format='fits', nside_io=4, hdf5_group='map'):
         """
         Write a HealSparseMap to a file.  Use the `metadata` property from
         the map to persist additional information in the fits header.
@@ -386,10 +385,8 @@ class HealSparseMap(object):
             File format.  May be ``fits``, ``parquet``, ``hdf5`` or ``healpix``. Note that
             the ``healpix`` EXPLICIT format does not maintain all metadata and
             coverage information.
-        group: `str`, optional
-            If file format is ``hdf5``, save to this group
-            otheriwse unused
-            default is ``map``
+        hdf5_group: `str`, optional
+            If file format is ``hdf5``, save to this group, otheriwse unused.
 
         Raises
         ------
@@ -397,7 +394,7 @@ class HealSparseMap(object):
         ValueError if nside_io is out of range.
         """
         _write_map(self, filename, clobber=clobber, nocompress=nocompress, format=format,
-                   nside_io=nside_io, hdf5_group=group)
+                   nside_io=nside_io, hdf5_group=hdf5_group)
 
     def write_moc(self, filename, clobber=False):
         """
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -162,7 +162,8 @@ class HealSparseMap(object):
         """
         return _read_map(cls, filename, nside_coverage=nside_coverage, pixels=pixels,
                          header=header, degrade_nside=degrade_nside,
-                         weightfile=weightfile, reduction=reduction, use_threads=use_threads, hdf5_group=group)
+                         weightfile=weightfile, reduction=reduction, use_threads=use_threads,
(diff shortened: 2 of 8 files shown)


healsparse/healSparseMap.py
```diff
@@ -149,6 +149,9 @@ def read(cls, filename, nside_coverage=None, pixels=None, header=False,
            Not yet implemented for parquet files.
         use_threads : `bool`, optional
            Use multithreaded reading for parquet files.
+        group : `str`, optional
+            if file is hdf5, read from this group
```
erykoff: -> `If file is hdf5, read from this group.` (Please format as a sentence).
The next line is not necessary; the default is in the signature.

healsparse/healSparseMap.py
```diff
@@ -379,17 +383,21 @@ def write(self, filename, clobber=False, nocompress=False, format='fits', nside_
             Must be less than or equal to nside_coverage, and not greater than 16.
             This option only applies if format=``parquet``.
         format : `str`, optional
-            File format.  May be ``fits``, ``parquet``, or ``healpix``. Note that
+            File format.  May be ``fits``, ``parquet``, ``hdf5`` or ``healpix``. Note that
             the ``healpix`` EXPLICIT format does not maintain all metadata and
             coverage information.
+        group: `str`, optional
+            If file format is ``hdf5``, save to this group
+            otheriwse unused
```
erykoff: -> `If file format is ``hdf5``, save to this group; otherwise unused.`
(Can fit on one line; next line is not necessary)

healsparse/healSparseMap.py
erykoff: Actually, looking further, I think that this should be `hdf_group` instead of `group` everywhere.

healsparse/healSparseMap.py
erykoff: As above, make this `hdf5_group`.

healsparse/healSparseMap.py
```diff
@@ -159,7 +162,8 @@ def read(cls, filename, nside_coverage=None, pixels=None, header=False,
         """
         return _read_map(cls, filename, nside_coverage=nside_coverage, pixels=pixels,
                          header=header, degrade_nside=degrade_nside,
-                         weightfile=weightfile, reduction=reduction, use_threads=use_threads)
+                         weightfile=weightfile, reduction=reduction, use_threads=use_threads,
+                         hdf5_group=group)
```
erykoff: -> `hdf_group=hdf5_group`.

healsparse/healSparseMap.py
```diff
@@ -379,17 +383,21 @@ def write(self, filename, clobber=False, nocompress=False, format='fits', nside_
...
+            default is ``map``
 
         Raises
         ------
         NotImplementedError if file format is not supported.
         ValueError if nside_io is out of range.
         """
         _write_map(self, filename, clobber=clobber, nocompress=nocompress, format=format,
-                   nside_io=nside_io)
+                   nside_io=nside_io, hdf5_group=group)
```
erykoff: -> `hdf_group=hdf5_group`.

healsparse/io_map.py
```diff
@@ -37,6 +38,8 @@ def _read_map(healsparse_class, filename, nside_coverage=None, pixels=None, head
         (mean, median, std, max, min, and, or, sum, prod, wmean).
     use_threads : `bool`, optional
         Use multithreaded reading for parquet files.
+    hdf5_group: `str`, optional
+        hdf5 group to read from (only used if reading from an hdf5 file)
```
erykoff: -> `Read from this hdf5 group (only used reading from an hdf5 file).`

healsparse/io_map.py
```diff
@@ -67,13 +74,18 @@ def _read_map(healsparse_class, filename, nside_coverage=None, pixels=None, head
                                  pixels=pixels, header=header, degrade_nside=degrade_nside,
                                  weightfile=weightfile, reduction=reduction,
                                  use_threads=use_threads)
+    elif is_hdf5_file:
+        return _read_map_hdf5(healsparse_class, filename, group=hdf5_group,
+                              pixels=pixels, header=header, degrade_nside=degrade_nside,
+                              weightfile=weightfile, reduction=reduction,)
```
erykoff: The trailing comma should only be there if using a true multi-line call.  So it should be removed here.

healsparse/io_map.py
```diff
@@ -67,13 +74,18 @@ def _read_map(healsparse_class, filename, nside_coverage=None, pixels=None, head
...
     elif not os.path.isfile(filename):
         raise IOError("Filename %s could not be found." % (filename))
     else:
-        raise NotImplementedError("HealSparse only supports fits and parquet files (with pyarrow).")
+        raise NotImplementedError("HealSparse only supports fits, hdf5 and parquet files (with pyarrow).")
```
erykoff: -> `... hdf5 (with h5py), and parquet ...`

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,252 @@
...
+def _write_map_hdf5(hsp_map, filepath, group="map", clobber=False):
+    """
+    Internal method to write a HealSparseMap to an HDF5 file in a specified group.
+
+    Supports regular, recarray, wide-mask, and bit-packed maps.
+
+    Parameters
+    ----------
+    hsp_map : HealSparseMap
+        Map to save
```
erykoff: -> `Map to save`.

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,252 @@
...
+    filepath : str
+        HDF5 file path
+    group : str, optional
+        Name of the HDF5 group to store the map (default 'map')
```
erykoff: No need to put the default in the docstring.  Here and below.

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,252 @@
...
+    clobber : bool, optional
+        Overwrite the file/group if it exists (default False)
+    """
+    if os.path.isfile(filepath) and not clobber:
+        raise RuntimeError("Filename %s exists and clobber is False." % (filepath))
```
erykoff: I'm not sure how this can work as is.  `clobber` is the same as `overwrite` which isn't happening here.  I think we can demand that the HDF5 already exists (document this!) and therefore (a) `clobber` does nothing and (b) we can demand that the file exists.

Or ... I see now that you're using `clobber` to overwrite an existing group?  In which case, that needs to be documented but the check here is *not* what you want.

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,252 @@
...
+
+        # Coverage map - only save valid pixels
+        coverage_pixels = np.where(hsp_map.coverage_mask)[0]
+        coverage_values = hsp_map.coverage_mask[coverage_pixels].astype(bool)
+        grp.create_dataset("coverage_pixel", data=coverage_pixels, compression="gzip")
+        grp.create_dataset("coverage_value", data=coverage_values, compression="gzip")
+
+        # Sparse map - only save valid pixels
+        valid_pixels = hsp_map.valid_pixels
+        grp.create_dataset("pixel", data=valid_pixels, compression="gzip")
```
erykoff: This is not going to work, and in fact may explode well beyond your existing memory.  The `healsparse` format is extremely efficient without recording pixel values (which can sometimes fail and is often slow).  You want to save the map itself, and not valid pixels.  Same thing with the coverage map.

erykoff (changes requested):
I've started to go through but not the full way.  The *map* should be saved and *not* the valid pixels.  This is a key part of the healsparse format.

healsparse/healSparseMap.py
elvinpoole: done

healsparse/healSparseMap.py
elvinpoole: done

healsparse/healSparseMap.py
elvinpoole: done

healsparse/healSparseMap.py
elvinpoole: done

healsparse/io_map.py
elvinpoole: done

healsparse/io_map.py
elvinpoole: done

healsparse/io_map.py
elvinpoole: done

healsparse/io_map_hdf5.py
elvinpoole: done

healsparse/io_map_hdf5.py
elvinpoole: done

healsparse/io_map_hdf5.py
elvinpoole: thanks. the clobber logic should now work. Yes, the intended behaviour is:
clobber=False only prevents overwriting the group. If the file already exists but the group is new, we can still save to a new group in that file (even if clobber=F)

healsparse/io_map_hdf5.py
elvinpoole: This should now save the map

healsparse/healSparseMap.py
```diff
@@ -379,17 +382,19 @@ def write(self, filename, clobber=False, nocompress=False, format='fits', nside_
             Must be less than or equal to nside_coverage, and not greater than 16.
             This option only applies if format=``parquet``.
         format : `str`, optional
-            File format.  May be ``fits``, ``parquet``, or ``healpix``. Note that
+            File format.  May be ``fits``, ``parquet``, ``hdf5`` or ``healpix``. Note that
             the ``healpix`` EXPLICIT format does not maintain all metadata and
             coverage information.
+        hdf5_group: `str`, optional
+            If file format is ``hdf5``, save to this group, otheriwse unused.
```
erykoff: -> otherwise

healsparse/io_map.py
```diff
@@ -47,17 +50,21 @@ def _read_map(healsparse_class, filename, nside_coverage=None, pixels=None, head
...
     is_fits_file = False
     is_parquet_file = False
+    is_hdf5_file = False
 
     try:
         fits = HealSparseFits(filename)
         is_fits_file = True
         fits.close()
-    except OSError:
+    except (OSError, UnicodeDecodeError):
```
erykoff, line 59: Where is the `UnicodeDecodeError` coming from?

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,308 @@
...
+        # re-shape sparse_map data so each coverage pixel is a different row
+        # chunk the dataset such that each chunk is 1 coverage pixel
+        ncov_in_sparse = sum(hsp_map.coverage_mask) + 1  # include buffer pixel
+        nfine_per_cov = hsp_map._cov_map._nfine_per_cov
+
+        if hsp_map.is_rec_array:
+            # for recarray, save each field separately
+            for name in hsp_map._sparse_map.dtype.names:
+                sparse_map_reshape = hsp_map[name]._sparse_map.reshape(ncov_in_sparse, nfine_per_cov)
+                field_grp = grp.create_group(name)
```
erykoff, line 61: So this will create basically `map.name`?

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,308 @@
...
+                field_grp.create_dataset(
+                    "sparse_map",
+                    data=sparse_map_reshape,
+                    chunks=(1, nfine_per_cov),
+                    compression="gzip",
+                )
+        elif hsp_map.is_bit_packed_map:
+            # save as bool array rather than packed to keep the same pixel as other maps when reading
+            sparse_map_reshape = np.asarray(hsp_map._sparse_map).reshape(ncov_in_sparse, nfine_per_cov)
```
erykoff: This is going to blow up the memory and crash in the cases that a packed map is being used.  Please use the packed format.  In this case you just need to reshape with a divide-by-8 on `nfine_per_cov` e.g. https://github.com/LSSTDESC/healsparse/blob/main/healsparse/io_map_fits.py#L614-L616

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,308 @@
...
+                chunks=(1, nfine_per_cov),
+                compression="gzip",
+                dtype=bool,
+            )
+
+        elif hsp_map.is_wide_mask_map:
+            # wide mask, save 2D values
+            sparse_map_reshape = hsp_map[name]._sparse_map.reshape(
+                ncov_in_sparse, nfine_per_cov, hsp_map.wide_mask_width
+            )
```
erykoff, line 83: Is this better than flattening each of the rows?  (It may not matter, but I'm curious).

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,308 @@
...
+        if is_rec_array:
+            dtype = []
+            for name in grp:
+                if name in ["cov_index_map"]:
+                    continue
+                dtype.append((name, grp[name]["sparse_map"].dtype))
+
+            sparse_map = np.zeros(sparse_size, dtype=dtype)
+            for name, _ in dtype:
+                sparse_map[name][:] = sentinel
```
erykoff: This isn't going to work because the sentinel only applies to `primary`.  You'll need to read the overflow coverage pixel in all cases (not just here)

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,308 @@
...
+                sparse_map[name] = grp[name]["sparse_map"][cov_index_in_sparse_ordered, :][inv].reshape(-1)
+        elif is_wide_mask:
+            sparse_map = (
+                grp["sparse_map"][cov_index_in_sparse_ordered, :][inv]
+                .reshape((-1, wide_mask_width))
+                .astype(WIDE_MASK)
+            )
+        elif is_bit_packed:
+            sparse_map = grp["sparse_map"][cov_index_in_sparse_ordered, :][inv].reshape(-1)
+            sparse_map = _PackedBoolArray.from_boolean_array(sparse_map)
```
erykoff: You'll need to leave this stored as packed.

tests/test_io_hdf5.py
```diff
@@ -0,0 +1,251 @@
...
+        sparse_map[30000:30005] = True
+
+        sparse_map.write(fname, format="hdf5")
+
+        sparse_map2 = healsparse.HealSparseMap.read(fname)
+
+        testing.assert_array_equal(sparse_map2[30000:30005], True)
+
+        self.assertEqual(len(sparse_map2.valid_pixels), 5)
+
```
erykoff, line 235: There needs to be a recarray test with mixed columns (e.g. different data types).

erykoff (changes requested):
Thanks!  This is looking a lot better but there are still a couple places that will not work in  general (packed bool maps and recarrays).  Also, please add to the documentation about this new format.

healsparse/io_map.py
elvinpoole: It seems this is what happens when you try to load the hdf5 file as if its fits. there might be a cleaner way of checking

healsparse/io_map_hdf5.py
elvinpoole: It saves each sparse_map in the recarray to a separate hdf5 group

healsparse/io_map_hdf5.py
elvinpoole: done

healsparse/io_map_hdf5.py
elvinpoole: I don't think it makes a difference. I only did it like this so that the 2nd dimension was still `nfine_per_cov` (same as the regular maps), seems neater that way. happy to change if there is a reason not to use 3D arrays

healsparse/io_map_hdf5.py
elvinpoole: i read the overflow pixel from file anyway so this line was doing nothing. removed

healsparse/io_map_hdf5.py
elvinpoole: done

tests/test_io_hdf5.py
elvinpoole: added

docs/filespec.rst
```diff
@@ -276,3 +276,123 @@ The other columns in the overflow coverage pixel are filled with the default sen
...
+
+.. _hdf5_format:
+
+HealSparseMap HDF5 Serialization
+================================
+
+A :code:`HealSparseMap` may also be serialized to an HDF5 file. 
+Multiple :code:`HealSparseMap` objects can be stored in the same HDF5 file.
+Each :code:`HealSparseMap` is stored in a different HDF5 group.
+They are not required to have the same mask, :code:`nside_coverage` or :code:`nside_sparse`.
```
erykoff: Oxford comma, please!

healsparse/healSparseMap.py
```diff
@@ -149,6 +149,8 @@ def read(cls, filename, nside_coverage=None, pixels=None, header=False,
            Not yet implemented for parquet files.
         use_threads : `bool`, optional
            Use multithreaded reading for parquet files.
+        hdf5_group : `str`, optional
+           If file format is ``hdf5``, read from this group, otheriwse unused.
```
erykoff: -> otherwise

healsparse/io_map.py
erykoff: Huh!  Okay, could you add a comment here?

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,307 @@
...
+                sparse_map_reshape = hsp_map[name]._sparse_map.reshape(ncov_in_sparse, nfine_per_cov)
+                field_grp = grp.create_group(name)
+                field_grp.create_dataset(
+                    "sparse_map",
+                    data=sparse_map_reshape,
+                    chunks=(1, nfine_per_cov),
+                    compression="gzip",
+                )
+        elif hsp_map.is_bit_packed_map:
+            # save as bool array rather than packed to keep the same pixel as other maps when reading
```
erykoff: This comment is no longer operative.

healsparse/io_map_hdf5.py
```diff
@@ -0,0 +1,307 @@
...
+    pixels : `list`, optional
+        List of coverage map pixels to read.
+    header : `bool`, optional
+        Return stored metadata/header as well as map?  Default is False.
+    degrade_nside : `int`, optional
+        Degrade map to this nside on read.
+    weightfile : `str`, optional
+        Weight map for weighted degrade.
+    reduction : `str`, optional
+        Reduction method for degrade-on-read.
```
erykoff, line 148: Should be a blank line before Returns

tests/test_io_hdf5.py
```diff
@@ -0,0 +1,310 @@
...
+from numpy import random
+import tempfile
+import shutil
+import os
+import pathlib
+
+import healsparse
+
+# read specific coverage pixels not yet implemented
+test_pixels_read = False
```
erykoff: Is this still true?

erykoff (commented):
A few small comments.  The only substantive one is that it looks like the "pixels read" tests are turned off, but it seems like these should work.  So these tests should be turned on (and working) or specifying pixels needs to raise a NotImplementedError.

healsparse/healSparseMap.py
elvinpoole: done

healsparse/io_map.py
elvinpoole: done

docs/filespec.rst
elvinpoole: done

healsparse/healSparseMap.py
elvinpoole: done

healsparse/io_map_hdf5.py
elvinpoole: removed

healsparse/io_map_hdf5.py
elvinpoole: done

tests/test_io_hdf5.py
elvinpoole: I removed this as an option and made the pixel tests always run

erykoff (approved):

PR #176, opened 2026-04-29
Inequality operators

elvinpoole:
- Added overloads of operators < <= > >= == and !=
- The == operator will revert to default behaviour if you try `hsp_map == hsp_map` and requires explicitly redefining the base object hash. I'm happy to modify or revert this one if there are any dangers in doing this
EDIT: we now dont re-define the hash and i've implemented the eq operator for map-map comparisons

### DIFF
--- a/healsparse/healSparseMap.py
+++ b/healsparse/healSparseMap.py
@@ -16,7 +16,6 @@ class HealSparseMap(object):
     """
     Class to define a HealSparseMap
     """
-
     def __init__(self, cov_map=None, cov_index_map=None, sparse_map=None, nside_sparse=None,
                  healpix_map=None, nside_coverage=None, primary=None, sentinel=None,
                  nest=True, metadata=None, _is_view=False):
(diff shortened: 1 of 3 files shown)


healsparse/healSparseMap.py
```diff
@@ -2358,7 +2420,7 @@ def __invert__(self):
             sentinel=self._sentinel,
         )
```
elvinpoole, line 2432: Note: this does expand the full sparse map, but I think the apply_operator function already did this with `self._sparse_map.copy()`

healsparse/healSparseMap.py
```diff
@@ -2374,6 +2446,9 @@ def _apply_operation(self, other, func, int_only=False, in_place=False):
             Only accept integer types.
         in_place : `bool`, optional
             Perform operation in-place.
+        sentinel : None, `int` or `float`
```
erykoff: I think this can be `int` or `float` or `bool`?

erykoff (approved):
This looks great, with a minor comment.

I can't believe that `_apply_operation()` works with this with only minor alterations!

healsparse/healSparseMap.py
elvinpoole: fixed

PR #177, opened 2026-07-01
Add support for rustfits fits backend.

erykoff:


healsparse/io_coverage_fits.py
```diff
@@ -25,7 +25,7 @@ def _read_coverage_fits(coverage_class, filename_or_fits):
 
     try:
         cov_index_map = fits.read_ext_data('COV')
-    except (OSError, KeyError):
+    except (OSError, KeyError, ValueError):
```
esheldon, line 32: What is the situation here?  It is a fits file but doesn't have the 'COV' extension?

healsparse/fits_shim.py
```diff
@@ -169,12 +205,14 @@ def ext_is_image(self, extension):
         is_image : `bool`
         """
         hdu = self.fits_object[extension]
-        if use_fitsio:
+        if use_rustfits:
+            return isinstance(hdu, (rustfits.CompressedImageHDU, rustfits.ImageHDU))
```
esheldon: You can just check `rustfits.ImageHDU`

healsparse/fits_shim.py
```diff
@@ -355,10 +389,18 @@ def _make_header(metadata, force_astropy=False):
     -------
     header : `fitsio.FITSHDR` or `astropy.io.fits.Header`
     """
+    if use_rustfits and not force_astropy:
+        if metadata is None:
+            return {}
+        else:
+            hdr = copy.copy(metadata)
+            for reserved in FITS_RESERVED:
```
esheldon, line 397: `rustfits` will ignore protected fits header keys when you write to a header from an existing `FITSHeader` using the `update` method, but not when writing from a `dict`, so it is correct to skip protected keys.

rustfits provides `rustfits.is_protected_key(name)` for the above, which could replace your check against `FITS_RESERVED`.

Note, if this metadata was itself created from a rustfits `FITSHeader`, you can do `header.to_dict(skip_protected=True)` to provide the clean dict.

It probably makes sense to provide a method to clean a dict of protected keys so you don't need the loop.

healsparse/io_map.py
```diff
@@ -56,9 +56,10 @@ def _read_map(healsparse_class, filename, nside_coverage=None, pixels=None, head
         fits = HealSparseFits(filename)
         is_fits_file = True
         fits.close()
-    except (OSError, UnicodeDecodeError):
+    except (OSError, UnicodeDecodeError, ValueError):
         pass
     # UnicodeDecodeError occurs when trying to read an hdf5 file as if it's FITS
+    # ValueError occurs from rustfits when trying to read an hdf5 file as if it's FITS
```
esheldon, line 62: What actual error is raised?  I'm guessing it is  ValueError because it tries to parse the primary header and finds the wrong values in the first bytes, but I want to see if a different error might be more appropriate to raise.

healsparse/fits_shim.py
erykoff: Because of the way that the tests do reading/writing, and there's no way to directly create a rustfits header in python, sometimes this code gets a rustfits header and sometimes a dict.  But I will make use of the `is_protected_key()`.

healsparse/io_map.py
erykoff: So the following code snippet raises `ValueError` because it can't access extension 0.  If the file `nothing.notfits` is not there then it raises an OSError (as expected).  If the file isn't a fits (e.g. `touch nothing.notfits`) then rustfits will happily "open" the file but won't raise (`ValueError`) until it tries to access an extension.  Should it be checking that it's a valid fits file on open?
```
with rustfits.FITS("nothing.notfits") as f:
    print(f[0])
```

healsparse/fits_shim.py
erykoff: Hmm.  I don't want to loop over all keys ... just the protected ones.  I just don't know how to do this without a bunch more special cases.

esheldon: There is no need to change what you are doing, I was just giving information, sorry to imply otherwise

Here is what rustfits does

https://github.com/esheldon/rustfits/blob/5c959d88ceaebefa3fd29cf4ee144adcd22e0757/src/header.rs#L440

healsparse/io_map.py
esheldon: rustfits.FITS should do eager header parsing on open.  It is true that fitsio was lazy

erykoff: But it doesn’t parse the header on an empty file?

esheldon: sounds like a bug, investigating

esheldon: this is fixed and will be in the next release

esheldon (approved):
I think this looks good.

I see there is the "append_extension" which uses rustfits the same way other tools would be used.  I think that's fine.

For heavy extending to compressed HDUs however, there is the better way, and healsparse should take advantage of it.  The use case is building a huge map in batches.  In that case the extending() context can be  used to get huge speedups (factor of ~20).  Obviously this can wait until the need arises.

PR #178, opened 2026-07-24
Use rustfits or astropy remote file access when possible.

erykoff:


healsparse/fits_shim.py
```diff
@@ -71,26 +80,55 @@ def __init__(self, filename, mode='r'):
...
+            self._use_rustfits = True
+
             if mode == "r":
                 rustfits_mode = "r"
             elif mode == "rw":
                 rustfits_mode = "r+"
 
-            self.fits_object = rustfits.FITS(filename, mode=rustfits_mode)
+            if mode == "r" and protocol != "file":
+                self.fits_object = rustfits.FITS(filename, mode=rustfits_mode, remote="ranged")
```
esheldon, line 98: Is this defaulting to always  using ranged?   I think often this is not the best choice, much faster to use the default "download all and work with it in memory"

setup.cfg
```diff
@@ -14,35 +14,38 @@ classifiers =
...
 test_with_healpy =
   healpy
 astropy =
-  astropy>=5.3
+  astropy>=6.0
 fitsio =
   fitsio>=1.4.0
 rustfits =
-  rustfits>=0.1.5
+  rustfits>=0.1.7
```
esheldon: I think ranged landed in 0.1.8

healsparse/fits_shim.py
erykoff: Hmm.  Part of the problem is that only rustfits supports this.  The other thing is that I was thinking that if somebody is handing a remote URI then they are asking to read it remotely (for good or ill).  I also worry about trying to get too many parameters in at the top and passing them all down ... but if you think it's necessary I could.

esheldon: what is only supported by rustfits?

erykoff: Right now I support various fits backends, but I am thinking that I should remove that ... or at least remove fitsio because I do require rustfits.  So I'll do that cleanup on this PR.

But I also support astropy fits reading which is currently required for remote access via S3 or google cloud or webdav (I don't think rustfits supports these or it would be a great surprise).  And only rustfits will "download the file to a temp cache and then read it".  I personally find that behavior surprising.  I would instead expect that if I give a remote protocol I want to read it remotely.

esheldon: Getting this to work took some real code and decisions, where as the download version didn't take any.  

Also the download is what cfitsio does.

From that point of view, ranged access not being default isn't a surprise, the ranged read existing at all is the nice addition (thanks for prompting it).  It is nice astropy does it too.

Why does S3, google cloud, webdave require anything other than authentication?  rustfits does support that through the keyword

esheldon: https://rustfits.readthedocs.io/en/latest/api/fits.html#remote-transport-configuration

PR #181, opened 2026-08-21
Fix index error when computing coverage_map and fracdet from maps with out-of-order coverage pixels

erykoff:
There was a bug where the fracdet map was garbage when the coverage pixels were out-of-order in memory.  This fixes the bug and also simplifies the code paths between the fracdet map and the coverage map.


beckermr (approved):
I cannot speak for the code changes, but the tests look correct. Lgtm!

